PR #11, opened 2024-04-25
feat(v3-sdk): add mixed quoter address on base

jsy1218:
## Description

We have deployed v2 everywhere. This means we should support mixed quote on every L2. We start by trying mixed quote on Base.

## How Has This Been Tested?

It will be tested in SOR and routing-api via the existing integ-tests. If this mixed quoter doesn't work, then the quote will throw unexpected error, and the tests will fail on Base.

## (Optional) Feedback Focus

No

## (Optional) Follow Ups

No


zhongeric (approved):
Lgtm but the name should be MixedRouteQuoterV1 I think to align with the contract name

zhongeric (approved):

adjkant:
@jsy1218 Confirming, this is intentionally a minor version (rather than a patch) bump correct?

marktoda (approved):

jsy1218:
> @jsy1218 Confirming, this is intentionally a minor version (rather than a patch) bump correct? 

Yes it's a minor version, because I consider adding mixed quoter on Base a backward-compatible new feature, not a bug fix. Prior to deploy v2 everywhere, mixed quoter on L2 was not a supported feature

PR #14, opened 2024-04-26
chore(release): fix breaking change marker

adjkant:
## Description

Due to rules being a glob, the `!` marker was treated as a negation and accidentally treated all minor version bumps as major. This removes that possibility!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,7 +6,7 @@ Please title your PR according to the following types and scopes following [conv
 - `chore(<type>):` will not trigger any release and should be used for internal repo changes
 - `<type>(public):` will trigger a patch version for non-code changes (e.g. README changes)
 - `feat(SDK name):` will trigger a minor version
-- `feat(!):` will trigger a major version for a breaking change
+- `feat(breaking):` will trigger a major version for a breaking change
 
 ## Description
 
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -6,7 +6,7 @@ Please title your PR according to the following types and scopes following [conv
 - `chore(<type>):` will not trigger any release and should be used for internal repo changes
 - `<type>(public):` will trigger a patch version for non-code changes (e.g. README changes)
 - `feat(SDK name):` will trigger a minor version
-- `feat(!):` will trigger a major version for a breaking change
+- `feat(breaking):` will trigger a major version for a breaking change
```
marktoda, line 9: does this scope to a specific SDK? if so how

ie can I do `feat(breaking sdk-core)`?

marktoda (approved):

jsy1218 (approved):

.github/pull_request_template.md
adjkant: The scope will need to be `breaking` literal, but which SDK it affects will only be based on the files changed! So changing for `sdk-core` would mean only touching files in that SDK for the PR with that title and `semantic-release` will handle automatically!

PR #30, opened 2024-05-23
feat(v3-sdk): export v3Swap logic

ewilz:
## Description

export v3 swap functionality so that it does not have to be copy-pasted for v4 sdk, which uses same swap logic (if a hook doesn't interfere)

## How Has This Been Tested?

N/A

No


sdks/v3-sdk/src/utils/swapMath.ts
```diff
@@ -17,7 +16,7 @@ export abstract class SwapMath {
     sqrtRatioTargetX96: JSBI,
     liquidity: JSBI,
     amountRemaining: JSBI,
-    feePips: FeeAmount
+    feePips: JSBI
```
ewilz, line 19: `FeeAmount` is an `enum` used in v3, v4 has no bounds on fees (besides an upper limit?), so I'm decoupling the enum from the swap functions.

adjkant: @ewilz Won't this be a breaking change?

adjkant (commented):
Looks good in terms of best practices/versions except for one comment!

sdks/v3-sdk/src/utils/swapMath.ts
```diff
@@ -26,12 +26,13 @@ export abstract class SwapMath {
       feeAmount: JSBI
     }> = {}
 
+    feePips = JSBI.BigInt(feePips)
```
adjkant, line 29: Does `JSBI.BigInt` work with JSBI input or should we do this conditionally based on the type?

ewilz: it will take a JSBI input (or else the tests would be failing actually)

sdks/v2-sdk/src/router.test.ts
```diff
@@ -28,7 +28,7 @@ describe('Router', () => {
 
   describe('#swapCallParameters', () => {
     describe('exact in', () => {
-      it.only('ether to token1', () => {
```
hensha256, line 31: lol nice

adjkant (approved):
Approving from a compat perspective, defer to protocols on the rest!

hensha256 (approved):
lgtm!

PR #31, opened 2024-05-23
feat(v4-sdk): initial commit

ewilz:
## PR Scope

## Description

Create v4-sdk

## How Has This Been Tested?

Unit tests, as this is a very minimal and contained sdk

 initial commit!

## (Optional) Feedback Focus

complies with v4 pool setups. Initial commit only works with v3 style pools. Things like dynamic fees or hook actions aren't considered for this sdk POC


sdks/v4-sdk/package.json
```diff
@@ -0,0 +1,96 @@
...
+  "scripts": {
+    "build": "tsdx build",
+    "lint": "tsdx lint src",
+    "release": "semantic-release",
+    "start": "tsdx watch",
+    "test": "tsdx test"
+  },
+  "dependencies": {
+    "@ethersproject/abi": "^5.5.0",
+    "@ethersproject/solidity": "^5.0.9",
```
jsy1218, line 28: nit: wondering if we can upgrade v4-sdk dependencies to ethers 6+.

(on the flip side, it will make merging all the sdks into a single npmjs artifact harder, since all existing sdks rely on ethers 5. So I can be persuaded otherwise)

sdks/v4-sdk/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "@uniswap/sdk-core": "^4.2.1-beta.1",
```
jsy1218, line 29: nit: sdk-core latest is 5.3.1

sdks/v4-sdk/.gitignore
```diff
@@ -0,0 +1,66 @@
...
+# parcel-bundler cache (https://parceljs.org/)
+.cache
+.parcel-cache
+
+# yarn v2
+.yarn/cache
+.yarn/unplugged
+.yarn/build-state.yml
+.yarn/install-state.gz
+.pnp.*
```
jsy1218, line 66: nit: add jetbrain webstorm artifact directories:

```
.idea
```

sdks/v4-sdk/src/entities/pool.test.ts
```diff
@@ -0,0 +1,303 @@
...
+  ADDRESS_ZERO,
+  FEE_AMOUNT_LOW,
+  FEE_AMOUNT_MEDIUM,
+  FEE_AMOUNT_HIGHEST,
+  NEGATIVE_ONE,
+  ONE_ETHER,
+  TICK_SPACING_TEN,
+} from '../utils/internalConstants'
+
+describe('Pool', () => {
```
jsy1218, line 15: we probably need some getPoolId unit test coverages.

sdks/v4-sdk/src/entities/pool.ts
```diff
@@ -0,0 +1,259 @@
...
+      'PRICE_BOUNDS'
+    )
+    // always create a copy of the list since we want the pool's tick list to be immutable
+    ;[this.currency0, this.currency1] = sortsBefore(currencyA, currencyB)
+      ? [currencyA, currencyB]
+      : [currencyB, currencyA]
+    this.fee = fee
+    this.sqrtRatioX96 = JSBI.BigInt(sqrtRatioX96)
+    this.tickSpacing = tickSpacing
+    this.hooks = hooks
```
jsy1218, line 93: nit: we probably want to validate hooks address during constructor, [checkValidAddress](https://github.com/Uniswap/sdks/blob/main/sdks/sdk-core/src/utils/validateAndParseAddress.ts#L22C17-L22C34).  I'm unsure about ethers library behavior in case of zero address though.

sdks/v4-sdk/src/entities/pool.ts
```diff
@@ -0,0 +1,259 @@
...
+  public readonly sqrtRatioX96: JSBI
+  public readonly hooks: string // address
+  public readonly liquidity: JSBI
+  public readonly tickCurrent: number
+  public readonly tickDataProvider: TickDataProvider
+
+  private _currency0Price?: Price<Currency, Currency>
+  private _currency1Price?: Price<Currency, Currency>
+
+  public static getPoolId(
```
jsy1218, line 37: we might get the benefit of consistency between client library and protocol contract, if we define a `PoolKey` type in v4-sdk. That way, getPoolId method parameters are also more flexible to add/remove new params without modifying the getPoolId function itself.

sdks/v4-sdk/package.json
```diff
@@ -0,0 +1,96 @@
...
+  ],
+  "license": "MIT",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "module": "dist/v4-sdk.esm.js",
+  "files": [
+    "dist"
+  ],
+  "engines": {
+    "node": ">=10"
```
jsy1218, line 17: nit: we can probably support node version much later than 10 for v4-sdk. ur-sdk supports >= 14, so we can probably do that for v4-sdk as well to start.

sdks/v4-sdk/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "lint": "tsdx lint src",
+    "release": "semantic-release",
+    "start": "tsdx watch",
+    "test": "tsdx test"
+  },
+  "dependencies": {
+    "@ethersproject/abi": "^5.5.0",
+    "@ethersproject/solidity": "^5.0.9",
+    "@uniswap/sdk-core": "^4.2.1-beta.1",
+    "@uniswap/v3-sdk": "3.12.0",
```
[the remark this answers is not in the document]
ewilz: I didn't mind exporting from v3, because v4 pools without any hooks ARE v3 pools (minus the singleton). They work in the same exact way.

sdks/v4-sdk/package.json
ewilz: this would make this package bigger, right? Since we'd be importing all of ethers?

sdks/v4-sdk/src/entities/pool.ts
```diff
@@ -0,0 +1,309 @@
...
+/**
+ * Represents a V4 pool
+ */
+export class Pool {
+  public readonly currency0: Currency
+  public readonly currency1: Currency
+  public readonly fee: number
+  public readonly tickSpacing: number
+  public readonly sqrtRatioX96: JSBI
+  public readonly hooks: string // address
```
marktoda, line 37: I wonder if we want some `Hooks` object to handle like parsing the flags

sdks/v4-sdk/src/entities/pool.ts
```diff
@@ -0,0 +1,309 @@
...
+  public readonly liquidity: JSBI
+  public readonly tickCurrent: number
+  public readonly tickDataProvider: TickDataProvider
+
+  // backwards compatibility with Pool interface v2/v3 sdk
+  public readonly token0: Currency
```
marktoda, line 43: Consider using getters to make these type aliases without having fully separate instances -- avoids potential for mis-setting it, i.e.

```
public get token0(): Currency {
   return this.currency0;
}
```

sdks/v4-sdk/src/entities/pool.test.ts
```diff
@@ -0,0 +1,368 @@
...
+          ADDRESS_ZERO,
+          encodeSqrtRatioX96(1, 1),
+          0,
+          0,
+          []
+        )
+      }).toThrow('FEE')
+    })
+
+    it('fee cannot be more than 1e6', () => {
```
ewilz, line 42: acutally I think this is not true for v4...

sdks/v4-sdk/src/entities/pool.ts
```diff
@@ -0,0 +1,309 @@
...
+   */
+  public get chainId(): number {
+    return this.currency0.chainId
+  }
+
+  public async getOutputAmount(
+    inputAmount: CurrencyAmount<Currency>,
+    sqrtPriceLimitX96?: JSBI
+  ): Promise<[CurrencyAmount<Currency>, Pool]> {
+    invariant(this.involvesCurrency(inputAmount.currency), 'CURRENCY')
```
marktoda, line 209: do we want something like "if the hook has beforeSwap or afterSwap, throw an error that we can't be sure"?

marktoda: ah nvm I see you have that in swap. Maybe a comment that clarifies this only gets output amount of the vanilla v3 swap piece?

marktoda (approved):

sdks/v4-sdk/src/entities/pool.ts
ewilz: that would be cool, but if getting this merged is important, can always be a followup task 🤷

marktoda (approved):

marktoda (approved):

sdks/v4-sdk/src/entities/pool.ts
```diff
@@ -0,0 +1,321 @@
...
+    return this.currency0
+  }
+  public get token1(): Currency {
+    return this.currency1
+  }
+
+  /**
+   * Returns true if the currency is either currency0 or currency1
+   * @param currency The currency to check
+   * @returns True if currency is either currency0 or currency
```
zhongeric, line 150: nit: comment

zhongeric (approved):
lgtm!

zhongeric (approved):

PR #46, opened 2024-06-17
feat(uniswapx-sdk): change MULTICALL_ADDRESS to be chainId specific

zhongeric:
## PR Scope

## Description

Since zksync has a different implementation of CREATE2, addresses are no longer consistent with other EVM chains. We've seen this in permit2 deployments. This PR updates the default exports of uniswapx-sdk to use the correct address for multicall3 on zksync, as per https://github.com/mds1/multicall/blob/main/deployments.json#L927

## How Has This Been Tested?

Tests

Any direct consumers of the `MULTICALL_ADDRESS` export of uniswapx-sdk will have to change to use `multicallAddressOn(chainId)`. I'm not aware of any codebases using this besides fee-collector, which is what motivated this change.

## (Optional) Feedback Focus

## (Optional) Follow Ups


codyborn (approved):
🚢

ConjunctiveNormalForm (approved):

PR #61, opened 2024-07-18
feat(router-sdk): Introduce V4 pools

ewilz:
## Description

_router-sdk needs to account for v4 pools. With this PR, integrators can construct trades and routes that involve V4 pools. However, you cannot encode routes with SwapRouter via this PR. That feature will be available later. _

## How Has This Been Tested?

_I still need to write the tests_

_no, I don't think so_

## (Optional) Feedback Focus

_Especially with the introduction of native currencies in V4, wrangling in token/currency types has been a challenge. pls tal if there's a better way to approach this,._

## (Optional) Follow Ups

_SwapRouter still cannot encode routes with V4 Pools. We don't do this in most places anyway. The only place SwapRouter is used for encoding is by Smart Order Router to use the quoter, but the current quoter can't accommodate V4 routes_


sdks/router-sdk/src/entities/mixedRoute/route.ts
```diff
@@ -13,9 +14,10 @@ type TPool = Pair | Pool
  */
 export class MixedRouteSDK<TInput extends Currency, TOutput extends Currency> {
   public readonly pools: TPool[]
-  public readonly path: Token[]
+  public readonly path: Currency[]
   public readonly input: TInput
```
jsy1218, line 18: we probably want to comment `input` field with @deprecated JSDoc. 

Ideally if we can change input from public to private, that'd be great, but it might break things. Or alternatively, we can remain `input` as public, but rename it to something like `input_deprecated` for awareness.

sdks/router-sdk/src/utils/encodeMixedRouteToPath.ts
```diff
@@ -1,37 +1,44 @@
...
+type TPool = Pair | V3Pool | V4Pool
 /**
  * Converts a route to a hex encoded path
  * @notice only supports exactIn route encodings
  * @param route the mixed path to convert to an encoded path
  * @returns the exactIn encoded path
  */
 export function encodeMixedRouteToPath(route: MixedRouteSDK<Currency, Currency>): string {
-  const firstInputToken: Token = route.input.wrapped
+  const firstInputToken: Currency = route.input.wrapped
```
jsy1218, line 17: route.adjustedInput?

sdks/router-sdk/src/entities/trade.ts
```diff
@@ -105,8 +122,8 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
     const poolAddressSet = new Set<string>()
     for (const { route } of this.swaps) {
       for (const pool of route.pools) {
-        if (pool instanceof Pool) {
-          poolAddressSet.add(Pool.getAddress(pool.token0, pool.token1, (pool as Pool).fee))
+        if (pool instanceof V3Pool) {
```
jsy1218, line 125: missing else if for V4Pool?

sdks/router-sdk/src/utils/encodeMixedRouteToPath.ts
```diff
@@ -1,37 +1,44 @@
...
     (
-      { inputToken, path, types }: { inputToken: Token; path: (string | number)[]; types: string[] },
-      pool: Pool | Pair,
+      { inputToken, path, types }: { inputToken: Currency; path: (string | number)[]; types: string[] },
+      pool: TPool,
       index
-    ): { inputToken: Token; path: (string | number)[]; types: string[] } => {
-      const outputToken: Token = pool.token0.equals(inputToken) ? pool.token1 : pool.token0
+    ): { inputToken: Currency; path: (string | number)[]; types: string[] } => {
+      if (pool instanceof V4Pool) throw 'Encoding mixed routes with V4 not supported'
```
jsy1218, line 25: nit: `throw new Error('Encoding mixed routes with V4 not supported')`

jsy1218 (approved):
Agree we should hold off on encoding routes. Personally, I don't really like the implementation of encoding routes, because on the quoter side, it has to decode the path to compute each pool address within the lens contract. It seems to me that with V4, we might have the opportunity to just pass in the PoolId in the correct order.

Let's hold off the route encoding for now, and we can have more discussions later.

sdks/router-sdk/src/entities/mixedRoute/route.ts
ewilz: the input field is still used

sdks/router-sdk/src/entities/mixedRoute/route.ts
```diff
@@ -13,9 +14,10 @@ type TPool = Pair | Pool
...
   public readonly output: TOutput
+  public readonly adjustedInput: Currency // routes with v2/v3 initial pool must wrap native input currency before trading
```
ewilz, line 20: could find a better name for this... this input isn't necessarily the input from the user, but the "adjusted" input for trading on the pool. i.e. the user sends ETH, but trades on a WETH pool first (so it must be wrapped)

sdks/router-sdk/src/entities/mixedRoute/trade.ts
```diff
@@ -48,7 +51,7 @@ export function tradeComparator<TInput extends Currency, TOutput extends Currenc
  * Represents a trade executed against a set of routes where some percentage of the input is
  * split across each route.
  *
- * Each route has its own set of pools. Pools can not be re-used across routes.
+ * Each route has its own set of pools. V3Pools can not be re-used across routes.
```
zhongeric, line 54: & V4Pools I assume?

sdks/router-sdk/src/entities/mixedRoute/trade.ts
```diff
@@ -193,19 +196,21 @@ export class MixedRouteTrade<TInput extends Currency, TOutput extends Currency,
...
-
     invariant(amount.currency.equals(route.input), 'INPUT')
-    amounts[0] = amount.wrapped
+
+    amounts[0] = route.pools[0] instanceof V4Pool ? amount : amount.wrapped
     for (let i = 0; i < route.path.length - 1; i++) {
       const pool = route.pools[i]
-      const [outputAmount] = await pool.getOutputAmount(amounts[i])
+      const [outputAmount] =
+        pool instanceof V4Pool ? await pool.getOutputAmount(amounts[i]) : await pool.getOutputAmount(amounts[i].wrapped)
```
zhongeric, line 210: Ah it's going to be quite difficult always having to do the instanceof check - could we write a helper function that calls `getOutputAmount` with either the raw currency or the wrapped version depending on the instance of pool passed in?

sdks/router-sdk/src/entities/mixedRoute/trade.ts
```diff
@@ -193,19 +196,21 @@ export class MixedRouteTrade<TInput extends Currency, TOutput extends Currency,
...
+    const amounts: CurrencyAmount<Currency>[] = new Array(route.path.length)
     let inputAmount: CurrencyAmount<TInput>
     let outputAmount: CurrencyAmount<TOutput>
 
     invariant(tradeType === TradeType.EXACT_INPUT, 'TRADE_TYPE')
-
     invariant(amount.currency.equals(route.input), 'INPUT')
-    amounts[0] = amount.wrapped
+
+    amounts[0] = route.pools[0] instanceof V4Pool ? amount : amount.wrapped
```
zhongeric, line 206: could we do route.adjustedInput here?

sdks/router-sdk/src/entities/mixedRoute/trade.ts
```diff
@@ -357,16 +365,20 @@ export class MixedRouteTrade<TInput extends Currency, TOutput extends Currency,
...
-        pool instanceof Pool
-          ? poolAddressSet.add(Pool.getAddress(pool.token0, pool.token1, pool.fee))
-          : poolAddressSet.add(Pair.getAddress(pool.token0, pool.token1))
+        if (pool instanceof Pair) {
+          poolIdentifierSet.add(Pair.getAddress(pool.token0, pool.token1))
+        } else if (pool instanceof V3Pool) {
+          poolIdentifierSet.add(V3Pool.getAddress(pool.token0, pool.token1, pool.fee))
+        } else if (pool instanceof V4Pool) {
+          poolIdentifierSet.add(pool.poolId)
+        }
```
zhongeric, line 377: I think we should throw in an else branch here for unknown pool types, would help catch mismatched package issues

sdks/router-sdk/src/swapRouter.ts
```diff
@@ -308,7 +310,7 @@ export abstract class SwapRouter {
             const exactInputParams = [
               i === 0 ? amountIn : 0, // amountIn
               !isLastSectionInRoute(i) ? 0 : amountOut, // amountOutMin
-              newRoute.path.map((token) => token.address), // path
+              newRoute.path.map((token) => token.wrapped.address), // path TODOEmily: this should only be wrapped sometimes
```
zhongeric, line 313: todo?

sdks/router-sdk/src/entities/mixedRoute/route.ts
zhongeric: ah yea, I don't mind the current naming, another idea could be `inputOrWrapped`?

zhongeric (commented):
looks good, just a few comments

sdks/router-sdk/src/entities/mixedRoute/trade.ts
ewilz: hahah woops this got caught erroneously in my find&replace

sdks/router-sdk/src/entities/mixedRoute/trade.ts
ewilz: I can throw this in, althoughhh I have noticed we have these kinds of checks exist in places, but realize they're redundant because every function that accepts pools has a `TPool = Pair | V3Pool | V4Pool`constraint for inputs

sdks/router-sdk/src/entities/mixedRoute/trade.ts
ewilz: it's an amount, not an input (which is currency only)

sdks/router-sdk/src/utils/getOutputAmount.ts
```diff
@@ -0,0 +1,20 @@
+import { Currency, CurrencyAmount } from '@uniswap/sdk-core'
+import { Pool as V4Pool } from '@uniswap/v4-sdk'
+import { Pair } from '@uniswap/v2-sdk'
+import { Pool as V3Pool } from '@uniswap/v3-sdk'
+
+type TPool = Pair | V3Pool | V4Pool
```
zhongeric, line 6: nit: this is defined in multiple places, can be shared

sdks/router-sdk/src/utils/getOutputAmount.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function getOutputAmount(
+  pool: TPool,
+  amountIn: CurrencyAmount<Currency>
+): Promise<[CurrencyAmount<Currency>, TPool]> {
+  if (pool instanceof V4Pool) {
+    if (pool.involvesCurrency(amountIn.currency)) {
+      return await pool.getOutputAmount(amountIn)
+    } else if (pool.involvesCurrency(amountIn.currency.wrapped)) {
```
zhongeric, line 15: I think this else if branch is redundant since the default behavior of this function is to return the outputAmount given the wrapped amountIn

zhongeric (approved):
looks good!

sdks/router-sdk/src/utils/getOutputAmount.ts
ewilz: ahh good point, I had a thrown error in there if it was neither, but realized it would never get hit and took it out.

PR #64, opened 2024-07-31
chore(infra): add provenance

mr-uniswap:
## Description

Adds provenance (a kind of code signing with an ephemeral key). Details in https://docs.npmjs.com/generating-provenance-statements

Unclear if our deploy (since it is a monorepo) will allow this code signing, we can revert if it fails (worst thing happens it just won't deploy rather than deploying something bad)


mr-uniswap:
Also note, I manually checked this, but there is an uncertainty on how this will work with a monorepo

marktoda (approved):

PR #73, opened 2024-08-08
feat(uniswapx-sdk): add overrides to multicall to support quoting priority orders

zhongeric:
## PR Scope

## Description

Priority orders are _invalid_ before they are valid. This breaks our existing assumptions during order quoting. This PR adds support for orders to define a specific blockOverride that should be applied at quote time, and adds support to the underlying multicall implementation to forward along this new field.

## How Has This Been Tested?

e2e tests

This may be breaking - initializing the quoter will now require a `JsonRpcProvider`. Additionally, consumers of the SDK should be aware that they must use a JsonRpcProvider which supports blockOverrides.


sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -265,12 +280,27 @@ export class UniswapXOrderQuoter
       return [order.order.serialize(), order.signature];
     });
 
+    // const orderParser = new UniswapXOrderParser();
+    // const priorityOrders = orders.filter((o) => orderParser.getOrderTypeFromEncoded(o.order.serialize(), o.order.chainId) === OrderType.Priority);
```
zhongeric, line 284: this is commented out for testing, because the integration tests don't deploy reactors via create2 and this will throw since the reactor address is technically unknown

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -155,6 +158,15 @@ async function checkTerminalStates(
           order.order.info.nonce
         );
         return cancelled ? OrderValidation.NonceUsed : OrderValidation.Expired;
+      }
+      // if the order has block overrides AND order validation is OK, it is invalid if current block number is < block override
+      else if (order.order.blockOverrides && validation === OrderValidation.OK) {
```
ConjunctiveNormalForm, line 163: nit: can `checkDefined(order.order.blockOverrides)` above this block to avoid having to do `!`

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -171,7 +183,7 @@ export class UniswapXOrderQuoter
   protected quoter: OrderQuoterContract;
 
   constructor(
-    protected provider: BaseProvider,
+    protected provider: JsonRpcProvider,
```
ConjunctiveNormalForm, line 186: nit: use `StaticJsonRpcProvider` maybe

ConjunctiveNormalForm (approved):
awesome work, ty!

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
zhongeric: oh looks like i dont even need the !

ConjunctiveNormalForm (approved):

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -155,6 +158,15 @@ async function checkTerminalStates(
           order.order.info.nonce
         );
         return cancelled ? OrderValidation.NonceUsed : OrderValidation.Expired;
+      }
+      // if the order has block overrides AND order validation is OK, it is invalid if current block number is < block override
+      else if (order.order.blockOverrides && validation === OrderValidation.OK) {
+        const blockNumber = await provider.getBlockNumber();
+        const blockOverrides = order.order.blockOverrides;
+        if(blockOverrides.number && blockNumber < parseInt(blockOverrides.number, 16)) {
```
marktoda, line 166: lint

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -155,6 +158,15 @@ async function checkTerminalStates(
...
+          return OrderValidation.OrderNotFillable;
```
marktoda, line 167: nit: NotFillableYet? To signify that it will be later?

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -275,12 +291,23 @@ export class UniswapXOrderQuoter
       return [order.order.serialize(), order.signature];
     });
 
+    // TODO: does not work with multiple different block overrides
+    const blockOverrides = orders.reduce((acc: BlockOverrides, order) => {
```
marktoda, line 295: the reduce here will just keep overwriting acc right? Maybe more clear way to describe this is with `Array.find(!!blockOverrides.number)`

marktoda: Or split into several multicalls, one for each block override

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
```diff
@@ -267,20 +278,39 @@ export class UniswapXOrderQuoter
...
 
   /// Get the results of a multicall for a given function
+  /// Each order with a blockOverride is multicalled separately
   private async getMulticallResults(
     functionName: string,
     orders: SignedOrder[]
   ): Promise<MulticallResult[]> {
-    const calls = orders.map((order) => {
-      return [order.order.serialize(), order.signature];
+    const ordersWithBlockOverrides = orders.filter((order) => order.order.blockOverrides);
```
marktoda, line 286: could group by ones with the same blockOverrides

marktoda (approved):

ConjunctiveNormalForm (approved):

sdks/uniswapx-sdk/src/utils/OrderQuoter.ts
zhongeric: that's a nice optimization, I'll do in follow up to unblock us

PR #81, opened 2024-08-28
fix(router-sdk): Address wrapped/native currency conditions in mixed routes with v4

ewilz:
## Description

Valid Routes:
- `v2/v3 --> v4 WETH unwraps to ETH`
- `v4 --> `v2/v3 ETH wraps to WETH`
Invalid Routes:
- `v4 --> v4 WETH unwraps to ETH`
- `v4 --> v4 ETH wraps to WETH`
v4 routes must trade through WETH/ETH pools to make this conversion

## How Has This Been Tested?

unit tests

I don't think so

## (Optional) Feedback Focus

make sure all conditions are met.

### DIFF
--- a/sdks/router-sdk/src/entities/mixedRoute/route.ts
+++ b/sdks/router-sdk/src/entities/mixedRoute/route.ts
@@ -4,6 +4,7 @@ import { Currency, Price, Token } from '@uniswap/sdk-core'
 import { Pair } from '@uniswap/v2-sdk'
 import { Pool as V3Pool } from '@uniswap/v3-sdk'
 import { Pool as V4Pool } from '@uniswap/v4-sdk'
+import { isValidTokenPath } from '../../utils/isValidTokenPath'
 
 type TPool = Pair | V3Pool | V4Pool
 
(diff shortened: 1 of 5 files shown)


sdks/router-sdk/src/utils/isValidTokenPath.ts
```diff
@@ -0,0 +1,25 @@
+import { Currency, Token } from '@uniswap/sdk-core'
+import { Pool as V4Pool } from '@uniswap/v4-sdk'
+import { Pair } from '@uniswap/v2-sdk'
+import { Pool as V3Pool } from '@uniswap/v3-sdk'
+
+type TPool = Pair | V3Pool | V4Pool
+
+export function isValidTokenPath(prevPool: TPool, currentPool: TPool, inputToken: Currency): boolean {
+  if (currentPool.involvesToken(inputToken as Token)) return true
```
zhongeric, line 9: this should always be true if inputToken is not native right, a comment could be helpful

zhongeric (approved):

sdks/router-sdk/src/utils/isValidTokenPath.ts
ewilz: I have to put `as Token` bc v2/v3 pools only accept token parameters..so I'm basically hacking typescript so it will compile. this will work if the inputToken IS native too....

sdks/router-sdk/src/entities/mixedRoute/route.ts
```diff
@@ -4,6 +4,7 @@ import { Currency, Price, Token } from '@uniswap/sdk-core'
 import { Pair } from '@uniswap/v2-sdk'
 import { Pool as V3Pool } from '@uniswap/v3-sdk'
 import { Pool as V4Pool } from '@uniswap/v4-sdk'
+import { isValidTokenPath } from '../../utils/isValidTokenPath'
 
 type TPool = Pair | V3Pool | V4Pool
```
jsy1218, line 9: nit: as I was coding v4 routing in SOR, I realized getting this type exported can be beneficial

jsy1218 (approved):

sdks/router-sdk/src/entities/mixedRoute/route.ts
ewilz: yah I also need to DRY and put this somewhere, I keep meaning to do taht

ewilz: I'm afraid of needing new approvals, so I'm just gonna merge..

lozanopo (approved):

PR #84, opened 2024-09-03
feat(uniswapx-sdk): Implement Dutch V3

alanhwu:
## Description

We are adding support for the new Dutch v3 orders in the UniswapX SDK

## How Has This Been Tested?

Unit testing for all files so far, will include integration after contracts are completely finalized

None. V3 was intentionally written in a way that guarantees V2 doesn't break. In the future, we should consider creating a CosignedDutch class that parents classes such as V2 and V3. This would require significant refactoring but would be cleaner.


sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+        { name: "swapper", type: "address" },
+        { name: "nonce", type: "uint256" },
+        { name: "deadline", type: "uint256" },
+        { name: "additionalValidationContract", type: "address" },
+        { name: "additionalValidationData", type: "bytes" },
+    ],
+    V3DutchInput: [
+        { name: "token", type: "address" },
+        { name: "startAmount", type: "uint256" },
+        { name: "curve", type: "V3Decay" },
```
zhongeric, line 74: i know the contract abi is changing still but the type name must match what is in the contract https://github.com/Uniswap/UniswapX/pull/269/files#diff-ba488278e700ac9fd3b42d4d36a035810f7f76a75de5b9fc7eedad3136bd2ccbR53 , so "NonlinearDutchDecay"

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+        return new UnsignedV3DutchOrder(
+            {
+                ...json,
+                nonce: BigNumber.from(json.nonce),
+                input: {
+                    ...json.input,
+                    startAmount: BigNumber.from(json.input.startAmount),
+                    curve: {
+                        relativeBlocks: json.input.curve.relativeBlocks,
+                        relativeAmounts: json.input.curve.relativeAmounts.map(BigNumber.from),
```
zhongeric, line 126: hmm i don't think BigNumber supports negative ints?

zhongeric: Might have to use BigInt or something

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+                this.hash(),
+                abiCoder.encode(
+                    [COSIGNER_DATA_TUPLE_ABI],
+                    [
+                        [
+                        cosignerData.decayStartBlock,
+                        cosignerData.exclusiveFiller,
+                        cosignerData.exclusivityOverrideBps,
+                        cosignerData.inputOverride,
+                        cosignerData.outputOverrides,
```
zhongeric, line 292: nit: format

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+            curve: {
+                relativeBlocks: decodedRelativeBlocks,
+                relativeAmounts,
+            },
+            maxAmount,
+        },
+        outputs: outputs.map(
+            ([token, startAmount,[relativeBlocks, relativeAmounts],recipient]: [
+                string, number, [BigNumber, number[]], string, boolean
+            ]) => ({
```
zhongeric, line 476: linter should fix this

sdks/uniswapx-sdk/src/order/index.ts
```diff
@@ -1,20 +1,30 @@
...
   | CosignedV2DutchOrder
+  | UnsignedV3DutchOrder
+  | CosignedV3DutchOrder
   | UnsignedPriorityOrder
   | CosignedPriorityOrder;
 
 export type Order = UniswapXOrder | RelayOrder;
+
+export function originalIfZero(value: BigNumber, original: BigNumber): BigNumber {
+  return value.isZero() ? original : value;
```
zhongeric, line 29: this should be in utils, not in this file

sdks/uniswapx-sdk/src/order/types.ts
```diff
@@ -114,3 +119,32 @@ export type PriorityInputJSON = Omit<
...
+
+export type V3DutchOutput = {
+  readonly token: string;
+  readonly startAmount: BigNumber;
+  readonly curve: V3Decay;
+  readonly recipient: string;
+};
+
+export type V3DutchOutputJSON = Omit<V3DutchOutput, "startAmount"> & {
+  startAmount: string;
```
zhongeric, line 149: don't you need to serialize the BigInt values in the decay curve too? For both input and output

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -0,0 +1,120 @@
+import { BigNumber } from "ethers";
+
+import { V3Decay } from "../order";
+/*
+These functions mimic the smart contract functions as closely as possible to ensure that the same results are produced.
+Essentially Solidity translated to TypeScript.
+*/
+/* eslint-disable */
```
zhongeric, line 8: nit: why disable? a bit hard to read

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -0,0 +1,120 @@
...
+      const elapsed = BigNumber.from(currentPoint - startPoint);
+      const duration = BigNumber.from(endPoint - startPoint);
+      if (endAmount.lt(startAmount)) {
+        return startAmount.sub(
+          startAmount.sub(endAmount.mul(elapsed).div(duration)) //muldivdown in contract
+        );
+      } else {
+        return startAmount.add(
+          endAmount.sub(startAmount.mul(elapsed).div(duration)) //muldivup in contract
+          //TODO: How can we do muldivup in JS?
```
zhongeric, line 88: Ah you don't have to, since BigNumber doesn't support floating numbers. The rounding up/down in solidity is handled internally by the math library. What you have here should be fine but not entirely sure. Best to write a test with real numbers and see what the output is

zhongeric (commented):
looking great so far!

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
```diff
@@ -0,0 +1,274 @@
...
+            Object.assign(this.getOrderInfo(), {
+                input: this.info.input,
+                outputs: this.info.outputs,
+                cosigner: this.info.cosigner,
+            }),
+            this.chainId,
+            this.permit2Address
+        );
+    }
+    //TODO: fromOrder()
```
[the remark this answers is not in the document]
alanhwu: implemented at the top of the file but forgot to remove reminder. fixed!

sdks/uniswapx-sdk/src/builder/V2DutchOrderBuilder.test.ts
```diff
@@ -574,7 +574,7 @@ describe("V2DutchOrderBuilder", () => {
   });
 
   describe("partial order tests", () => {
-    it("builds an unsigned partial order with default cosignerData values", () => {
+    it("builds an unsigned partial order with default cosignerData values", () => { //TODO: partial orders don't have cosignerData...
```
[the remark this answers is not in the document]
alanhwu: My understanding was that `buildPartial()` returns the type `UnsignedV3DutchOrder` which has a `cosigner` field for address, but doesn't actually include a `cosignerData` field to house values so it's unable to be assigned defaults

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+import { BigNumber, ethers } from "ethers";
+
+import { getPermit2, ResolvedUniswapXOrder } from "../utils";
+import { getBlockDecayedAmount } from "../utils/dutchBlockDecay";
+
+import { BlockOverrides, OffChainOrder, OrderInfo, V3DutchInput, V3DutchInputJSON, V3DutchOutput, V3DutchOutputJSON, V3OrderResolutionOptions } from "./types";
+
+import { originalIfZero } from ".";
+
+export type CosignerDataJSON = {
```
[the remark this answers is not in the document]
alanhwu: renamed!

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+}
+
+function encodeRelativeBlocks(relativeBlocks: number[]): BigNumber {
+    let packedData = BigNumber.from(0);
+    for (let i = 0; i < relativeBlocks.length; i++) {
+        packedData = packedData.or(BigNumber.from(relativeBlocks[i]).shl(i * 16));
+    }
+    return packedData;
+}
+/* eslint-disable */
```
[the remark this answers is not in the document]
alanhwu: fixed

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,514 @@
...
+                [relativeBlocks, relativeAmounts],
+                maxAmount,
+            ],
+            outputs,
+            [decayStartBlock, exclusiveFiller, exclusivityOverrideBps, inputOverride, outputOverrides],
+            cosignature,
+        ],
+    ] = decoded;
+
+    const decodedRelativeBlocks = decodeRelativeBlocks(relativeBlocks);
```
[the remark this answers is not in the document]
alanhwu: Good catch, it's not guaranteed. This was handled but at different times between input vs output `relativeBlocks` which made it confusing with variable names and call patterns. Refactored, lmk what you think!

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: The linter wants `next` and `prev` inside of `locateArrayPosition` to be declared as `const` even though they are both reassigned in the while-loop. While I could rewrite the function to an equivalent implementation, I intentionally want to preserve identical code to the contract so that there aren't any edge-case differences in math. 

Agreed that disabling linter for the whole file was too far. Removed and only included directives for the 2 lines above `next` and `prev`!

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
alanhwu: updated!

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: Wrote a test for the rounding up to show the difference that we'd be looking at [here](https://github.com/Uniswap/sdks/blob/d63c65e325b329631e50e94a4b8adc571a80f20c/sdks/uniswapx-sdk/src/utils/dutchBlockDecay.test.ts#L23).
We don't get a round-up in the current implementation, but also want to double-check with @codyborn if the test case reflects the contract's behavior.

sdks/uniswapx-sdk/src/utils/order.ts
```diff
@@ -147,3 +158,19 @@ export class RelayOrderParser extends OrderParser {
     return RelayOrder.parse(order, chainId);
   }
 }
+
+export function isCosignedV2(
```
codyborn, line 162: WDYT about having both `UnsignedV2DutchOrder` and `UnsignedV3DutchOrder` implement an interface that has a function `isCosigned()`? Then we can run `order.isCosigned()` and it know how to perform the check.

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
```diff
@@ -0,0 +1,273 @@
...
+        };
+    }
+
+    input(input: V3DutchInput): this {
+        this.info.input = input;
+        return this;
+    }
+
+    output(output: V3DutchOutput): this {
+        invariant(
```
codyborn, line 156: flagging this in case you want to remove it

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,515 @@
...
+    outputOverrides: string[];
+}
+
+export type UnsignedV3DutchOrderInfoJSON = Omit<UnsignedV3DutchOrderInfo, "nonce" | "input" | "outputs" | "cosignerData"> & {
+    nonce: string;
+    input: V3DutchInputJSON;
+    outputs: V3DutchOutputJSON[];
+};
+
+export type V3CosignerData = {
```
codyborn, line 25:
```suggestion
export type V3CosignerData = Omit<CosignerData, "decayStartTime" | "decayEndTime" > & 
 {
```

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,515 @@
...
+    cosignature: string;
+};
+
+type V3WitnessInfo = {
+    info: OrderInfo,
+    cosigner: string,
+    baseInput: V3DutchInput,
+    baseOutputs: V3DutchOutput[],
+};
+
```
codyborn, line 51: nit: space

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -0,0 +1,515 @@
...
+        packedData = packedData.or(BigNumber.from(relativeBlocks[i]).shl(i * 16));
+    }
+    return packedData;
+}
+
+function decodeRelativeBlocks(packedData: BigNumber): number[] {
+    const relativeBlocks: number[] = [];
+    for (let i = 0; i < 16; i++) {
+        const block = packedData.shr(i * 16).toNumber() & 0xFFFF;
+        if (block !== 0) {
```
codyborn, line 510: Why is this `0` check here? Idk why we _would_ put a `0` in the relativeBlocks but skipping `0` values seems strange.

codyborn: Maybe instead, we can accept the array length as an argument if this is intended to determine the end.

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
alanhwu: Fixed. This was huge, thanks for catching!

sdks/uniswapx-sdk/src/trade/V3DutchOrderTrade.ts
```diff
@@ -0,0 +1,159 @@
...
+  public get outputAmount(): CurrencyAmount<TOutput> {
+    return this.getFirstNonFeeOutputStartEndAmounts().startAmount;
+  }
+
+  public minimumAmountOut(): CurrencyAmount<TOutput> {
+    return this.getFirstNonFeeOutputStartEndAmounts().endAmount;
+  }
+
+  public maximumAmountIn(): CurrencyAmount<TInput> {
+    const endAmount = getEndAmount({
```
codyborn, line 120: We shouldn't assume that the last value is the max input.

codyborn: Same for `minimumAmountOut()`

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -0,0 +1,124 @@
...
+	curve: NonlinearDutchDecay,
+	targetValue: number
+): [number, number] {
+	const relativeBlocks = curve.relativeBlocks;
+	// eslint-disable-next-line prefer-const
+	let prev = 0;
+	// eslint-disable-next-line prefer-const
+	let next = 0;
+
+	while (next < curve.relativeAmounts.length) {
```
codyborn, line 18: Just a heads up, per Eric's rec I've updated the contract to use a for loop instead.

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -0,0 +1,124 @@
...
+		if (relativeBlocks[next] >= targetValue) {
+			return [prev, next];
```
codyborn, line 20: Nit: there's some extra indentation in this file

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -0,0 +1,124 @@
...
+		const elapsed = BigNumber.from(currentPoint - startPoint);
+		const duration = BigNumber.from(endPoint - startPoint);
+		if (endAmount.lt(startAmount)) {
+			return startAmount.sub(
+				(startAmount.sub(endAmount)).mul(elapsed).div(duration) //muldivdown in contract
+			);
+		} else {
+			return startAmount.add(
+				(endAmount.sub(startAmount)).mul(elapsed).div(duration) //muldivup in contract
+				//TODO: How can we do muldivup in JS?
```
codyborn, line 93: muldivup just rounds up the result.. in the latest PR I've gone with `muldivdown` for both:
https://github.com/Uniswap/UniswapX/blob/c0d699883b02668aed1d95077efb4a78ec7d4cfc/src/lib/DutchDecayLib.sol#L94

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -0,0 +1,124 @@
...
+	const { decayStartBlock, startAmount, relativeBlocks, relativeAmounts } = config;
+	return NonLinearDutchDecayLib.decay({ relativeAmounts, relativeBlocks }, startAmount, decayStartBlock, atBlock);
+}
+
+export function getEndAmount(
+	config: Partial<DutchBlockDecayConfig>
+): BigNumber {
+	const { startAmount, relativeAmounts } = config;
+	if (!startAmount || !relativeAmounts) {
+		throw new Error("Invalid config for getting V3 decay end amount"); //TODO: Should we throw?
```
codyborn, line 121: I think throwing the error here makes sense to catch dev errors quickly since this should never be the case.

sdks/uniswapx-sdk/src/order/types.ts
alanhwu: fixed, serialized them to strings for JSON

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
alanhwu: removed and added test for it, ty!

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: Ty for your input! Will keep it to throw then

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: updated to align!

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: great, this will eliminate an edge case. updated!

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
alanhwu: Indeed it was intended to determine the end. If we add an invariant to not allow `0` elements, it'd allow us to easily use `0` to determine the end.

Accepting array length could accomplish this too, but you could also just scan all the bytes to find out how many elements there are without needing outside info passed in.

An edge case that comes to mind with `0` is that it could potentially make sense if it's the first element `relativeBlocks`.
Consider:
`relativeBlocks = [0,2,4]`
`relativeAmounts = [10, 20, 40]`
`startAmount = 100`
This would imply that the order has an **effective** `startAmount` of `110`

But allowing `0` elements to exist also allows for ambiguous cases like `relativeBlocks = [0,0,0,2]` to exist. And allowing it only for the first element feels like a unnecessary edge case to support.

Because of this, I think we should:
Add an invariant to enforce no `0` elements and continue to use it as the delimiter for our encoding
Could also consider an addition in the contract to `revert InvalidDecayCurve()` if we detect this.

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: No longer a concern since the contract was updated to use `mulDivDown` for both cases!

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
alanhwu: No longer a concern after the new implementation of this functionality in the contract!

sdks/uniswapx-sdk/src/utils/order.ts
alanhwu: I see how we can do this more cleanly. I think in the future an interface would be useful if we do a larger refactoring of Dutch V2 & V3 to share more code together. For now since we're just interested in accessing cosigned status, I replaced the 2 functions with a single one that uses generics. Lmk what you think!

sdks/uniswapx-sdk/src/trade/V3DutchOrderTrade.ts
alanhwu: Rewrote these implementations with extra tests [here](https://github.com/Uniswap/sdks/pull/84/commits/438218183e3a4f47c428c0752f6d4dbe19592d83)!

sdks/uniswapx-sdk/src/builder/V2DutchOrderBuilder.ts
```diff
@@ -287,9 +287,9 @@ export class V2DutchOrderBuilder extends OrderBuilder {
       "exclusivityOverrideBps not set"
     );
     invariant(
-      this.info.cosignerData.inputOverride !== undefined &&
+      this.info.cosignerData.inputOverride !== undefined && // inputOverride is defaulted to 0 because enforced to be of type BigNumber
```
codyborn, line 290: nit: can you move the comment to its own line instead of inline?

If inputOverride is defaulted to `0`, why do we check if it's not `undefined`?

alanhwu: This is a `V2` file where I spotted this mistake and was just making a note of it since I noticed this condition doesn't do anything. I will go ahead and remove it.

sdks/uniswapx-sdk/src/trade/V3DutchOrderTrade.ts
```diff
@@ -0,0 +1,126 @@
...
+		const relativeAmounts: bigint[] = nonFeeOutput.curve.relativeAmounts;
+		const startAmount: BigNumber = nonFeeOutput.startAmount;
+		// Get the maximum of the relative amounts
+		const maxRelativeAmount = relativeAmounts.reduce((max, amount) => amount > max ? amount : max, BigInt(0));
+		// minimum is the start - the max of the relative amounts
+		const minOut = startAmount.sub(maxRelativeAmount.toString());
+		return CurrencyAmount.fromRawAmount(this.outputAmount.currency, minOut.toString());
+	}
+
+	public maximumAmountIn(): CurrencyAmount<TInput> {
```
codyborn, line 89: It'd be nice to have this be a derived field similar to the `minimumAmountOut` above (instead of having the caller calculate it separately.

codyborn (commented):
Looks great! Left some small comments.

sdks/uniswapx-sdk/src/utils/order.ts
alanhwu: Though of a clean way to just use `instanceof`!

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    private checkUnsignedInvariants(info: Partial<CosignedV3DutchOrderInfo>): asserts info is UnsignedV3DutchOrderInfo {
+        invariant(info.cosigner !== undefined, "cosigner not set");
+        invariant(info.input !== undefined, "input not set");
+        invariant(
+            info.outputs && info.outputs.length > 0,
+            "outputs not set"
+        );
+        // Check if input curve is valid
+        invariant(info.input.curve.relativeAmounts.length === info.input.curve.relativeBlocks.length, "relativeBlocks and relativeAmounts length mismatch");
```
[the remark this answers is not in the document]
alanhwu: That'll be enforced by typings since `input` and `output` were checked to be set right before

sdks/uniswapx-sdk/src/trade/V3DutchOrderTrade.ts
alanhwu: added a helper function for this!

sdks/uniswapx-sdk/src/constants.ts
```diff
@@ -94,6 +95,7 @@ export const REACTOR_ADDRESS_MAPPING: ReactorMapping = {
     [OrderType.Dutch_V2]: "0x1bd1aAdc9E230626C44a139d7E70d842749351eb",
     [OrderType.Dutch]: "0x0000000000000000000000000000000000000000",
     [OrderType.Relay]: "0x0000000000000000000000000000000000000000",
+    [OrderType.Dutch_V3]: "0x0000000000000000000000000000000000000000",
```
codyborn, line 98: Let's update this to a deployed address before we merge it.

alanhwu: done!

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
```diff
@@ -0,0 +1,339 @@
...
+      this.permit2Address
+    );
+  }
+
+  // A helper function for users of the class to easily the value to pass to maxAmount in an input
+  static getMaxAmountOut(
+    startAmount: BigNumber,
+    relativeAmounts: bigint[]
+  ): BigNumber {
+    if (relativeAmounts.length == 0) {
```
codyborn, line 327: I think we should support the empty curve for limit orders.

codyborn: Let's add a test for this too

sdks/uniswapx-sdk/src/trade/V3DutchOrderTrade.ts
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	public minimumAmountOut(): CurrencyAmount<TOutput> {
+		const nonFeeOutput: V3DutchOutput = this.order.info.outputs[0];
+		const relativeAmounts: bigint[] = nonFeeOutput.curve.relativeAmounts;
+		const startAmount: BigNumber = nonFeeOutput.startAmount;
+		// Get the maximum of the relative amounts
+		const maxRelativeAmount = relativeAmounts.reduce((max, amount) => amount > max ? amount : max, BigInt(0));
+		// minimum is the start - the max of the relative amounts
+		const minOut = startAmount.sub(maxRelativeAmount.toString());
```
codyborn, line 85: Same here

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
alanhwu: Fixed

sdks/uniswapx-sdk/src/trade/V3DutchOrderTrade.ts
alanhwu: Since this one is max, I was able to use `BigInt(0)` as a base so even if the array is empty or all negative, the min will be defaulted to 0. So when it adds at the end it will correctly give us `startAmount`

sdks/uniswapx-sdk/src/builder/V3DutchOrderBuilder.ts
```diff
@@ -325,13 +325,20 @@ export class V3DutchOrderBuilder extends OrderBuilder {
...
   ): BigNumber {
     if (relativeAmounts.length == 0) {
-      throw new Error("relativeAmounts cannot be empty");
+      return startAmount;
     }
+
     // Find the minimum of the relative amounts
     const minRelativeAmount = relativeAmounts.reduce(
       (min, amount) => (amount < min ? amount : min),
       relativeAmounts[0]
```
codyborn, line 334: it might be cleaner to set this initial value as `0`  and then you can remove lines 336-341 (since subtracting 0 will give you the startAmount).

alanhwu: Fixed

codyborn (approved):
Nice job!

PR #87, opened 2024-09-05
feat(v4-sdk): Add V4Planner to encode v4 routes and positions

ewilz:


sdks/v4-sdk/src/utils/encodeRouteToPath.ts
```diff
@@ -0,0 +1,16 @@
+export const encodeRouteToPath = (poolKeys: any[], currencyIn: string): any[] => {
```
jsy1218, line 1: aha, I actually implemented v4 route `encodeRouteToPath` in SOR: https://github.com/Uniswap/smart-order-router/blob/main/src/providers/on-chain-quote-provider.ts#L444. Feel free to copy and paste here, so that I can leverage your work.

jsy1218: and it's already working, as I wrote an quote test in alpha route integ test for a v4 pool.

sdks/v4-sdk/src/utils/v4Planner.ts
```diff
@@ -0,0 +1,131 @@
...
+    const actionType = trade.tradeType === TradeType.EXACT_INPUT ? Actions.SWAP_EXACT_IN : Actions.SWAP_EXACT_OUT
+    const inputCurrency = trade.inputAmount.currency
+
+    for (let swap of trade.swaps) {
+      this.addAction(actionType, [
+        {
+          inputCurrency.isNative ? ADDRESS_ZERO : inputCurrency.wrapped.address,
+          path: encodeRouteToPath(swap.route),
+          amountIn: swap.inputAmount,
+          amountOutMinimum: /* need to do slippage checks for each individual route??? */,
```
ewilz, line 110: @zhongeric this is what kinda stopped me in my tracks for a second

zhongeric: Ah yes so this is a bit tricky - looking at the UR sdk code, for every `swap` in the trade, it generates a new Trade from that route, and calculates the amountOutMinimum for each subTrade within the larger Trade.

I don't think we need to do any of this swap/liquidity specific logic here in the planner. The classes can be in charge of adding the correct Actions in the correct order IMO

ewilz: Heyy, not exactly sure what you mean by your second statement here. Since everything that needs to happen in a V4 Trade is encapsulated within an Action that's submitted to a general V4Command, I wanted to abstract that away completely into the V4 planner for separation of concerns etc etc. so I think I will include a slippage check arg here. 

let me know if I misunderstood what you're saying, if you have an alternative suggestion!

sdks/v4-sdk/src/utils/encodeRouteToPath.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const encodeRouteToPath = (route: Route<Currency, Currency>): PathKey[] => {
+  let currencyIn = route.input
+  let pathKeys: PathKey[] = []
+
+  for (let pool of route.pools) {
+    const currencyOut = currencyIn.equals(pool.currency0) ? pool.currency1 : pool.currency0
+
+    pathKeys.push({
+      intermediateCurrency: currencyOut.isNative ? ADDRESS_ZERO : currencyOut.address,
```
jsy1218, line 21: we do need to pass in `exactOutput`, just like v3 `encodeRouteToPath` input function params. The reason is because in case of exact out, intermediateCurrency will be currencyIn for current for-loop iteration.

jsy1218 (approved):
left one comment, otherwise lgtm

sdks/v4-sdk/src/utils/v4Planner.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export type RouterAction = {
+  action: Actions
+  encodedInput: string
+}
+
+export function createAction(action: Actions, parameters: any[]): RouterAction {
+  const encodedInput = defaultAbiCoder.encode(ABI_DEFINITION[action], parameters)
+  return { action, encodedInput }
+}
```
zhongeric, line 156: do we need this if we can just use the planner?

zhongeric (approved):
lgtm! +1 to siyu's comment though

sdks/v4-sdk/src/utils/v4Planner.ts
zhongeric: ah yeah

zhongeric (approved):

zhongeric (approved):

PR #88, opened 2024-09-06
feat(breaking): UR address change for v4

dannydaniil:
## PR Scope

## Description

- This PR aims to introduce UR configs according to it's version
- It purposefully introduces a breaking change with the new required parameter so integrators are aware they're moving to a new router that requires new approvals etc etc
- Fixes API-437

## How Has This Been Tested?

Yes, `UNIVERSAL_ROUTER_ADDRESS` and `UNIVERSAL_ROUTER_CREATION_BLOCK`

## (Optional) Feedback Focus

## (Optional) Follow Ups


ewilz:
Can we do something like:

```
export enum RouterVersion {
  V1_2,
  V2,
}

export const latestRouterVersion() {
    return RouterVersion.V2
}

export const UNIVERSAL_ROUTER_ADDRESS = (version: RouterVersion, chainId: number): string => {
....
} 
```

sdks/universal-router-sdk/src/utils/constants.ts
```diff
@@ -1,8 +1,8 @@
 import { BigNumber } from 'ethers'
 
 type ChainConfig = {
-  v3Router: string
-  v4Router: string
+  v1_2Router: string
```
ewilz, line 4: we'll probably eventually have to do something similar with creationBlock. prefer alpha order based on thing, not version?? hence `routerV1_2`, `routerV2_0`? then we can have creationBlockV-whatever down the line

dannydaniil: great callout!

jsy1218 (approved):

sdks/universal-router-sdk/src/utils/constants.ts
```diff
@@ -1,129 +1,285 @@
...
+      [RouterVersion.V2_0]: {
+        address: '0x28731BCC616B5f51dD52CF2e4dF0E78dD1136C06',
+        creationBlock: 12640979,
+      },
+    },
   },
 }
 
-export const UNIVERSAL_ROUTER_ADDRESS = (chainId: number): string => {
+// Todo: Change `CHAIN_CONFIGS` to pull the UR address with v4
```
ewilz, line 273: should the Todo on line 20 suffice? These seem noisy. I don't think this code will change here

dannydaniil: addressed!

ewilz (approved):

PR #90, opened 2024-09-06
feat(v4-sdk): Add V4 PositionManager SDK skeleton

zhongeric:
## Description

Adds a skeleton implementation of the Position Manager for V4. It also adds the Position entity, which is basically copied over from V3 but with some type changes to allow for the use of native currencies. For the full implementation, see #86 

And enumerating through the cases we want to cover:
Create pool -> createCallParameters
Create pool, mint position (with liquidity) -> addCallParameters with flags
Mint position with existing pool -> addCallParameters with flags
Add liquidity to existing position in existing pool -> addCallParameters
Decrease liquidity from position and burn the position -> removeCallParameters with flags
Decrease liquidity from position -> removeCallParameters
Collect fees from position -> collectCallParameters

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/v4-sdk/src/posm.ts
```diff
@@ -0,0 +1,354 @@
...
+  /**
+   * The account that should receive the tokens.
+   */
+  recipient: string
+}
+
+export type MintOptions = CommonOptions & CommonAddLiquidityOptions & MintSpecificOptions
+export type IncreaseOptions = CommonOptions & CommonAddLiquidityOptions & ModifyPositionSpecificOptions
+
+export type DecreaseLiquidityOptions = CommonOptions & RemoveLiquiditySpecificOptions & ModifyPositionSpecificOptions
```
snreynolds, line 131: BurnLiquidityOptions?

sdks/v4-sdk/src/posm.ts
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  /**
+   * Produces the calldata for completely or partially exiting a position
+   * @param position The position to exit
+   * @param options Additional information necessary for generating the calldata
+   * @returns The call parameters
+   */
+  public static removeCallParameters(_position: Position, _options: DecreaseLiquidityOptions): MethodParameters {
+    // Encode DECREASE_LIQUIDITY and then COLLECT
```
snreynolds, line 205: this will be either a decrease or burn and then an encoding of take pair

sdks/v4-sdk/src/posm.ts
```diff
@@ -0,0 +1,354 @@
...
+  expectedCurrencyOwed1: CurrencyAmount<Currency>
+
+  /**
+   * The account that should receive the tokens.
+   */
+  recipient: string
+}
+
+export type MintOptions = CommonOptions & CommonAddLiquidityOptions & MintSpecificOptions
+export type IncreaseOptions = CommonOptions & CommonAddLiquidityOptions & ModifyPositionSpecificOptions
```
snreynolds, line 129: I think CommonAddLiquidityOptions could share the slippageTolerance param?

Is that percentage applied to each amount then? (amount0 and amount1) ?

sdks/v4-sdk/src/posm.ts
snreynolds, line 131: Where is the decrease/collect slippageTolerance param?

snreynolds (commented):
are you pulling the utils from v3-sdk?

and then I think we can separately have a PR for encoding all the actions with a v4 base actions planner?

sdks/v4-sdk/src/posm.ts
zhongeric: yeah i think its applied to each amount in V3

sdks/v4-sdk/src/posm.ts
zhongeric: Ah added to CommonOptions

zhongeric:
I think we should implement the actions router in #87 , which can be separate from this PR which doesn't do any implementation work

sdks/v4-sdk/src/posm.ts
snreynolds: Sorry I just meant are there going to be specific ones for burn? You could use RemoveLiquiditySpecificOptions I guess but there is no need for a liquidityPercentage to be passed in

sdks/v4-sdk/src/entities/position.ts
```diff
@@ -0,0 +1,194 @@
...
+   * Returns the price of token0 at the upper tick
+   */
+  public get token0PriceUpper(): Price<Currency, Currency> {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * Returns the amount of token0 that this position's liquidity could be burned for at the current pool price
+   */
+  public get amount0(): CurrencyAmount<Currency> {
```
snreynolds, line 73: doesn't this calculation need the current price and price at lower and price at upper?

snreynolds: discussed offline - will do similar to v3 impl.

sdks/v4-sdk/src/posm.ts
snreynolds: under the hood though if a user wants to burn we should just encode a BURN call not a decrease THEN burn

sdks/v4-sdk/src/utils/abi.ts
```diff
@@ -0,0 +1,524 @@
+// TODO: import this from npm
```
snreynolds, line 1: where would you publish this?

sdks/v4-sdk/src/utils/actions.ts
```diff
@@ -0,0 +1,75 @@
+import { defaultAbiCoder } from 'ethers/lib/utils'
+
```
snreynolds, line 2: would be nice to keep router + posm actions using the same `encodeAction` code so whenever that gets merged maybe we re architect this so we have posm actions, router actions and then they share a "planner" file that does the encoding/building of actions

zhongeric: yea i agree, ok we can wait on #87 for this one then. Talked to Ayo and they're ok with that

snreynolds (approved):
nice!

sdks/v4-sdk/src/utils/abi.ts
zhongeric: hmmm I'm not entirely sure but the v3-periphery abis are published in npm

sdks/v4-sdk/src/utils/actions.ts
zhongeric: ok actually going to remove this to unblock them

sdks/v4-sdk/src/entities/position.ts
```diff
@@ -0,0 +1,189 @@
...
+    amount1,
+    useFullPrecision,
+  }: {
+    pool: Pool
+    tickLower: number
+    tickUpper: number
+    amount0: BigintIsh
+    amount1: BigintIsh
+    useFullPrecision: boolean
+  }) {
```
dannydaniil, line 136: could we please add return types for this one and below?

dannydaniil (commented):
looks great, just added 1 comment

snreynolds (approved):
looks good to me

dannydaniil (approved):
lgtm

PR #91, opened 2024-09-08
chore(infra): Adds secret scanning

mr-uniswap:
## PR Scope

Adds secret scanning

## Description
Adds secret scanning on precommit hooks and at the PR (just in case you add --no-verify) 

## How Has This Been Tested?

Manually

No

## (Optional) Feedback Focus

Make sure that the install is still easy for users on the read me. I reran it just to be safe, but it's a case of "it works on my machine"

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,13 +22,14 @@
   "packageManager": "yarn@3.2.3",
   "private": true,
   "scripts": {
+    "g:install": "yarn install && husky",
     "g:check:deps:mismatch": "manypkg check",
     "g:build": "turbo run build",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
   "packageManager": "yarn@3.2.3",
   "private": true,
   "scripts": {
+    "g:install": "yarn install && husky",
```
adjkant, line 25: Can we do this with `postinstall` instead?

mr-uniswap: Good idea!

adjkant (approved):

PR #93, opened 2024-09-10
fix(universal-router-sdk): only do aggregated slippage check for trades with multiple routes

ewilz:


zhongeric (approved):

sdks/universal-router-sdk/src/entities/protocols/uniswap.ts
```diff
@@ -187,7 +187,8 @@ function addV2Swap<TInput extends Currency, TOutput extends Currency>(
       // if native, we have to unwrap so keep in the router for now
       routerMustCustody ? ROUTER_AS_RECIPIENT : options.recipient,
       trade.maximumAmountIn(options.slippageTolerance).quotient.toString(),
-      trade.minimumAmountOut(options.slippageTolerance).quotient.toString(),
+      // if router will custody funds, we do aggregated slippage check from rotuer
```
hensha256, line 190:
```suggestion
      // if router will custody funds, we do aggregated slippage check from router
```

hensha256 (approved):

PR #98, opened 2024-09-16
feat(v4-sdk): Create and Add Implementation

snreynolds:
## PR Scope

## Description

This PR will just implement `createCallParameters` and `addCallParameters` to reduce the size of the PR. All permit2 forwarding is also out of scope in this PR.

Note that @zhongeric has started the impl for remove and collect but I will be removing them from this PR and finalizing those implementations in a different PR.

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/v4-sdk/src/utils/currencyMap.ts
```diff
@@ -0,0 +1,10 @@
+import { Currency } from '@uniswap/sdk-core'
+import { ADDRESS_ZERO } from '../internalConstants'
+
+// Uniswap v4 supports native pools. Those currencies are represented by the zero address.
+// TODO: Figure out if this is how we should be handling weird edge case tokens like CELO/Polygon/etc..
```
snreynolds, line 5: I believe this has been resolved and that any Currency's isNative value will be set according to how it should be "treated" in the interface

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -155,13 +159,16 @@ export type RemoveLiquidityOptions = CommonOptions & RemoveLiquiditySpecificOpti
 export type CollectOptions = CommonOptions & CollectSpecificOptions
 
 // type guard
-// @ts-ignore
 function isMint(options: AddLiquidityOptions): options is MintOptions {
   return Object.keys(options).some((k) => k === 'recipient')
 }
 
+function shouldCreatePool(options: MintOptions): boolean {
+  return options.createPool === true && options.sqrtPriceX96 !== undefined
```
codyborn, line 167: Should we throw an exception if `createPool` is true but  `sqrtPriceX96` is undefined? Ie. is that likely a dev error?

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -178,150 +185,85 @@ export abstract class V4PositionManager {
...
-    const calldatas = ['0x', '0x']
-    let value = toHex(0)
+    invariant(JSBI.greaterThan(position.liquidity, ZERO), 'ZERO_LIQUIDITY')
 
-    return {
-      calldata: Multicall.encodeMulticall(calldatas),
-      value,
-    }
-  }
+    const calldatas: string[] = []
```
codyborn, line 199: nit: given that "data" is already plural, maybe we can name it "calldataSet" or "calldataList"

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -178,150 +185,85 @@ export abstract class V4PositionManager {
...
-    return {
-      calldata: Multicall.encodeMulticall(calldatas),
-      value: toHex(0),
+    // need to settle both currencies when minting / adding liquidity
+    planner.addSettlePair(position.pool.currency0, position.pool.currency1)
+
+    // Any sweeping must happen after the settling.
+    let value: string = toHex(0)
+    if (options.useNative) {
+      invariant(position.pool.currency0.isNative || position.pool.currency1.isNative, 'NO_NATIVE')
```
codyborn, line 241: Do we have a constants file to move `NO_NATIVE` into?

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -178,150 +185,85 @@ export abstract class V4PositionManager {
...
-    throw new Error('not implemented')
-  }
-
-  // @ts-ignore
-  private static encodeTakePair(currency0: Currency, currency1: Currency, recipient: string): string {
-    throw new Error('not implemented')
+    return V4PositionManager.INTERFACE.encodeFunctionData('initializePool', [
+      poolKey,
+      sqrtPriceX96.toString(),
+      hookData ?? '0x',
```
codyborn, line 262: nit: 'initializePool' and '0x' to constants

sdks/v4-sdk/src/entities/position.ts
```diff
@@ -52,59 +52,262 @@ export class Position {
...
-    throw new Error('Not implemented')
+    return tickToPrice(this.pool.currency0, this.pool.currency1, this.tickUpper)
   }
 
   /**
    * Returns the amount of token0 that this position's liquidity could be burned for at the current pool price
    */
   public get amount0(): CurrencyAmount<Currency> {
-    throw new Error('Not implemented')
+    if (this._token0Amount === null) {
```
codyborn, line 69: `if (!this._token0Amount)` will include the case it's `undefined`

sdks/v4-sdk/src/entities/position.ts
```diff
@@ -52,59 +52,262 @@ export class Position {
...
+      if (this.pool.tickCurrent < this.tickLower) {
+        this._token0Amount = CurrencyAmount.fromRawAmount(
+          this.pool.currency0,
+          SqrtPriceMath.getAmount0Delta(
+            TickMath.getSqrtRatioAtTick(this.tickLower),
+            TickMath.getSqrtRatioAtTick(this.tickUpper),
```
codyborn, line 75: just double checking.. this should be `tickLower` and `tickUpper` if it's less than the `tickLower`?

sdks/v4-sdk/src/PositionManager.ts
snreynolds: hm, yes! we should I think. if you want to create a pool you MUST specify a price

sdks/v4-sdk/src/PositionManager.ts
snreynolds: yeahh thats true, I think this was a pattern we just use elsewhere but I agree with you so I changed it to calldataList

sdks/v4-sdk/src/entities/position.ts
snreynolds: oh got it cool! updated. out of curiousity, could it ever be set to undefined instead of null ?

sdks/v4-sdk/src/entities/position.ts
snreynolds: yep! good question.

So on L70 we check where the tickCurrent is. 

If the tickCurrent is below the tickLower  (tickCurrent < tickLower) that means the full range of liquidity is above the current price. I think about it like this in my head:

```

__________|______________|||||||__
          ^              ^     ^
__________tC____________tL_____tU
```

Anything below tC (tickCurrent) is going to be entirely in amount1 and anything above tickCurrent is going to be entirely in amount0. 

Since this function wants to know what the amount0 is, then in this case we need to query for the amount between the tickLower and the tickUpper. Hopefully that makes sense!

sdks/v4-sdk/src/PositionManager.ts
snreynolds: moved!

sdks/v4-sdk/src/entities/position.ts
codyborn: Oh no, you're right.. I see it being initialized to null.

Ayoakala (commented):
Looks great so far and thanks for cranking this out! Left a few comments and asks but the SDK contract looks very similar to V3 which is helpful

sdks/v4-sdk/src/entities/position.ts
codyborn: I don't think I fully understand. If amount0 is everything to the right of `tC`, wouldn't it be between `tC` and `tL`?

sdks/v4-sdk/src/utils/v4PositionPlanner.ts
```diff
@@ -0,0 +1,92 @@
...
+  // MINT_POSITION
+  addMint(
+    pool: Pool,
+    tickLower: number,
+    tickUpper: number,
+    liquidity: BigintIsh,
+    amount0Max: BigintIsh,
+    amount1Max: BigintIsh,
+    owner: string,
+    hookData?: string
```
codyborn, line 19:
```suggestion
    hookData: string = EMPTY_BYTES
```

sdks/v4-sdk/src/utils/v4PositionPlanner.ts
```diff
@@ -0,0 +1,92 @@
...
+    const inputs = [toAddress(currency0), toAddress(currency1)]
+    this.addAction(Actions.SETTLE_PAIR, inputs)
+  }
+
+  // TAKE_PAIR
+  addTakePair(currency0: Currency, currency1: Currency, recipient: string): void {
+    const inputs = [toAddress(currency0), toAddress(currency1), recipient]
+    this.addAction(Actions.TAKE_PAIR, inputs)
+  }
+
```
codyborn, line 87: nit: this one is missing a comment

sdks/v4-sdk/src/PositionManager.ts
codyborn: '0x' -> EMPTY_BYTES

sdks/v4-sdk/src/multicall.ts
```diff
@@ -9,12 +9,14 @@ export abstract class Multicall {
...
-      calldatas = [calldatas]
+  public static encodeMulticall(calldataList: string | string[]): string {
+    if (!Array.isArray(calldataList)) {
+      calldataList = [calldataList]
     }
 
-    return calldatas.length === 1 ? calldatas[0] : Multicall.INTERFACE.encodeFunctionData('multicall', [calldatas])
+    return calldataList.length === 1
+      ? calldataList[0]
+      : Multicall.INTERFACE.encodeFunctionData('multicall', [calldataList])
```
codyborn, line 19: Since `calldataList` is already an array, do we need to wrap it again in an array?

codyborn (approved):
Looks good! Left small comments

sdks/v4-sdk/src/entities/position.ts
snreynolds: ah yeah so everything to the right of tickCurrent is ALL of amount0 on the pool, since we are just working with a specific position (who is defined from tL to tU) we just care about the amount0 between tL and tU.

snreynolds: if we wanted to know amount1 (where the state of the pool is the same in the above graph.. ie the position is entirely above the current tick) then amount1 = 0 since the position does not touch any of amount1. And you can verify this logic in the amount1() function call

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -155,13 +159,20 @@ export type RemoveLiquidityOptions = CommonOptions & RemoveLiquiditySpecificOpti
...
 
 // type guard
-// @ts-ignore
 function isMint(options: AddLiquidityOptions): options is MintOptions {
   return Object.keys(options).some((k) => k === 'recipient')
 }
 
+function shouldCreatePool(options: MintOptions): boolean {
+  if (options.createPool) {
+    invariant(options.sqrtPriceX96 !== undefined, 'NO_SQRT_PRICE')
```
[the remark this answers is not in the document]
snreynolds: Hm do you prefer it the other way? I feel like this makes more sense to me because the pool on the position object shouldnt exist if createPool = true. Like its passed in because you're now initializing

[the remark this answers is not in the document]
snreynolds: i mean it could exist but I feel like its more error prone bc in all other cases you should initialize a position with a pool that has all current up to date values

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -178,150 +189,85 @@ export abstract class V4PositionManager {
...
+    planner.addSettlePair(position.pool.currency0, position.pool.currency1)
+
+    // Any sweeping must happen after the settling.
+    let value: string = toHex(0)
+    if (options.useNative) {
+      invariant(position.pool.currency0.isNative || position.pool.currency1.isNative, NO_NATIVE)
+      let nativeCurrency: Currency = position.pool.currency0.isNative
+        ? position.pool.currency0
+        : position.pool.currency1
+      value = position.pool.currency0.isNative ? toHex(amount0Max) : toHex(amount1Max)
```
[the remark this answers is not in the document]
snreynolds: what do you mean rather than inputtedAmount? Can you link me to the line in v3?

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -178,150 +189,85 @@ export abstract class V4PositionManager {
     }
   }
 
-  public static addCallParameters(_position: Position, _options: AddLiquidityOptions): MethodParameters {
+  // TODO: Add Support for permit2 batch forwarding
+  public static addCallParameters(position: Position, options: AddLiquidityOptions): MethodParameters {
```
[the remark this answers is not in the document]
snreynolds: thats in a separate here ! https://github.com/Uniswap/sdks/pull/102

I broke out the PRs bc this one was getting to be very large

sdks/v4-sdk/src/utils/abi.ts
```diff
@@ -0,0 +1,524 @@
+// TODO: import this from npm
+export const abi = [
```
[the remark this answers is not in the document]
snreynolds: i think this is the most up to date abi yes, but let me know if you see discrepancies!

sdks/v4-sdk/src/utils/priceTickConversions.ts
```diff
@@ -0,0 +1,54 @@
...
+    ? new Price(baseCurrency, quoteCurrency, Q192, ratioX192)
+    : new Price(baseCurrency, quoteCurrency, ratioX192, Q192)
+}
+
+/**
+ * Returns the first tick for which the given price is greater than or equal to the tick price
+ * @param price for which to return the closest tick that represents a price less than or equal to the input price,
+ * i.e. the price of the returned tick is less than or equal to the input price
+ */
+export function priceToClosestTick(price: Price<Currency, Currency>): number {
```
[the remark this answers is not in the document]
snreynolds: it would be the exact same as v3 so I think you can just use the same function from v3-sdk

sdks/v4-sdk/src/utils/priceTickConversions.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!price.lessThan(nextTickPrice)) {
+      tick++
+    }
+  } else {
+    if (!price.greaterThan(nextTickPrice)) {
+      tick++
+    }
+  }
+  return tick
+}
```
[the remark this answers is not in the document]
snreynolds: same for this, I'd say just use the one from v3-sdk. it would be the exact same code

sdks/v4-sdk/src/PositionManager.ts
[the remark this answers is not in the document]
Ayoakala: Yep its fine with me; was just calling out for my own sake

sdks/v4-sdk/src/PositionManager.ts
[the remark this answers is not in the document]
Ayoakala: Yea was referring to [this](https://github.com/Uniswap/v3-sdk/blob/main/src/nonfungiblePositionManager.ts#L270)

We use `amountDesired` (inputtedAmount) instead of `amountMax `

sdks/v4-sdk/src/multicall.ts
zhongeric: because the mutlicall signature requires [], and the underlying functions being called in periphery require a list of actions (calldataList)

sdks/v4-sdk/src/PositionManager.ts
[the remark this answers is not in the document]
zhongeric: I think this is desired behavior because unlike in v3 we bound increasing positions by the maximum tokens required, so we need to send the MAX, and sweep any excess back to the user. This allows the price to slip "up" to your slippage.

codyborn (approved):
🚢

zhongeric (approved):

PR #101, opened 2024-09-20
feat: add priority order trade

marktoda:
This commit adds a priority order trade. Note that given the nature of
priority orders, the output amounts are all the `minimum`, since the
actual output amount is determined dynamically based on the priority fee
of the fill

## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/uniswapx-sdk/src/trade/PriorityOrderTrade.ts
```diff
@@ -0,0 +1,154 @@
...
+      );
+
+    this._firstNonFeeOutputAmount = amount;
+    return amount;
+  }
+
+  // TODO: revise when there are actually multiple output amounts. for now, assume only one non-fee output at a time
+  public get outputAmount(): CurrencyAmount<TOutput> {
+    // TODO: estimate epected amount, using classic quote or expected priority
+    return this.getFirstNonFeeOutputAmount();
```
ConjunctiveNormalForm, line 111: given amount in quote.info.outputs (assuming exact_in trade) is classic adjusted for gas , we can theoretically calculate the classic price here. Agree it's better to be able to pass in classic price, which requires a URA change (to add that info to the quote response) and was what I was planning to do

ConjunctiveNormalForm (approved):

sdks/uniswapx-sdk/src/trade/PriorityOrderTrade.ts
marktoda: Ah true, that won't be precise tho right since gas could be different from quote to quote?

codyborn (approved):

PR #102, opened 2024-09-20
feat(v4-sdk): support removeCallParameters

snreynolds:
## PR Scope

## Description

This PR depends on https://github.com/Uniswap/sdks/pull/98. I think they should be merged separately though.

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -96,11 +105,6 @@ export interface RemoveLiquiditySpecificOptions {
    * The optional permit of the token ID being exited, in case the exit transaction is being sent by an account that does not own the NFT
    */
   permit?: NFTPermitOptions
-
-  /**
-   * Parameters to be passed on to collect
-   */
-  collectOptions: Omit<CollectSpecificOptions, 'tokenId'>
```
Ayoakala, line 103: Hmm curious why we remove this? do we not collect fees when we remove in v4?

Ayoakala (commented):
Looks good so far! Remains Collect and we should be done right?

sdks/v4-sdk/src/PositionManager.ts
zhongeric: Fees are auto collected when you decrease & take together

zhongeric (approved):

zhongeric (approved):

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -189,6 +193,7 @@ export abstract class V4PositionManager {
     }
   }
 
+  // TODO: Add Support for permit2 batch forwarding
```
hensha256, line 196: would probably be better to open an issue or track instead of TODOs in the code?

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -257,6 +262,66 @@ export abstract class V4PositionManager {
...
+  /**
+   * Produces the calldata for completely or partially exiting a position
+   * @param position The position to exit
+   * @param options Additional information necessary for generating the calldata
+   * @returns The call parameters
+   */
+  public static removeCallParameters(position: Position, options: RemoveLiquidityOptions): MethodParameters {
+    /**
+     * cases:
+     * - if liquidityPercentage is 100%, encode BURN_POSITION and then TAKE_PAIR
```
hensha256, line 274: this doesnt actually seem to be the logic its following?
Its not doing `if options.liquidityPercentage.equalTo(ONE) -> burn, else -> decrease`

hensha256 (approved):

sdks/v4-sdk/src/PositionManager.ts
zhongeric: yes agree, there's already an open PR for this #104 but will do for future

PR #104, opened 2024-09-23
feat(v4-sdk): add permit2 forwarder support

zhongeric:
## PR Scope

## Description

Relies on #102 

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups

### DIFF
--- a/sdks/v4-sdk/src/PositionManager.ts
+++ b/sdks/v4-sdk/src/PositionManager.ts
@@ -77,14 +77,19 @@ export interface CommonAddLiquidityOptions {
   useNative?: NativeCurrency
 
   /**
-   * The optional permit parameters for spending token0
+   * The optional ERC2612 permit parameters for approving token0 for Permit2
    */
-  token0Permit?: any // TODO: add permit2 permit type here
+  token0Permit?: any 
 
   /**
-   * The optional permit parameters for spending token1
+   * The optional ERC2612 permit parameters for approving token1 for Permit2
    */
-  token1Permit?: any // TODO: add permit2 permit type here
+  token1Permit?: any 
+
+  /**
+   * The optional permit2 batch permit parameters for spending token0 and token1
+   */
+  batchPermit?: BatchPermitOptions
 }
 
 /**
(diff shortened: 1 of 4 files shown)


sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -205,6 +236,25 @@ export abstract class V4PositionManager {
...
+    // We use permit2 to approve tokens to the position manager
+    if (options.batchPermit) {
+      calldataList.push(
+        V4PositionManager.encodePermitBatch(
+          options.batchPermit.owner,
+          options.batchPermit.permitBatch,
+          options.batchPermit.signature
+        )
+      )
+    }
```
zhongeric, line 256: I feel like we don't need to support permit2 permit for either token0 OR token1. We can just do a batchPermit of 1 element in that case, simplifying the options interface

sdks/v4-sdk/src/PositionManager.test.ts
```diff
@@ -8,6 +8,7 @@ import {
   SQRT_PRICE_1_1,
   TICK_SPACINGS,
   ZERO_LIQUIDITY,
+  PositionFunctions,
```
hensha256, line 11: why are the test files stored in `src`? is there a reason?

sdks/v4-sdk/src/PositionManager.test.ts
```diff
@@ -43,12 +45,15 @@ describe('PositionManager', () => {
     []
   )
 
-  const recipient = '0x0000000000000000000000000000000000000003'
-
   const tokenId = 1
   const slippageTolerance = new Percent(1, 100)
   const deadline = 123
 
+  const mockOwner = '0x0000000000000000000000000000000000000001'
```
hensha256, line 52: This is the same address as currency0... probably fine but a bit weird? would rather they were different

hensha256: oh yeah especially as the mock owner is signing over currency0 in the tests

sdks/v4-sdk/src/utils/ERC2612Permit.ts
```diff
@@ -0,0 +1,40 @@
...
+  v: 0 | 1 | 27 | 28
+  r: string
+  s: string
+  owner: string
+  spender: string
+  amount: BigintIsh
+  deadline: BigintIsh
+}
+
+const IERC2612PermitAbi = [
```
hensha256, line 14: v4 doesnt support this 🥲

sdks/v4-sdk/src/PositionManager.test.ts
zhongeric: yea not sure, this is the pattern from v3-sdk

sdks/v4-sdk/src/PositionManager.test.ts
zhongeric: ah yeah good point

sdks/v4-sdk/src/PositionManager.test.ts
hensha256: weirddd

sdks/v4-sdk/src/PositionManager.ts
hensha256: i wonder how much more gas it is. But yeah we can always do this as a follow-up? Maybe worth opening an issue or something just in case its way more gas

hensha256 (approved):

sdks/v4-sdk/src/PositionManager.ts
zhongeric: Yep just made an issue #107

PR #115, opened 2024-09-27
Fix sig verification

codyborn:
`recoverAddress()` will hash the message first before recoverying. Since `signMessage()` does the same (hash first) then they will produce the same address.


Fart10 (approved):
..

Fart10 (approved):
.

PR #135, opened 2024-10-01
fix(router-sdk): Prepare router-sdk to fully support mixed routes

ewilz:


jsy1218 (approved):

sdks/router-sdk/src/utils/pathCurrency.ts
```diff
@@ -0,0 +1,35 @@
...
+      return pool.token0
+    } else if (pool.token1.wrapped.equals(currency)) {
+      return pool.token1
+    }
+
+    // otherwise the token is invalid
+  } else {
+    throw new Error(`Expected currency ${currency.symbol} to be either ${pool.token0.symbol} or ${pool.token1.symbol}`)
+  }
+  return currency.wrapped
```
ewilz, line 34: this is so typescript doesn't yell at me 🥱 this line should never get hit!!! But I don't know what else to do.....

jsy1218 (approved):

sdks/router-sdk/src/utils/pathCurrency.ts
```diff
@@ -0,0 +1,36 @@
...
+    return currency
+
+    // return if currency.wrapped if pool involves wrapped currency
+  } else if (pool.involvesToken(currency.wrapped as Token)) {
+    return currency.wrapped
+
+    // return native currency if pool involves native version of wrapped currency (only applies to V4)
+  } else if (pool instanceof V4Pool) {
+    if (pool.currency0.wrapped.equals(currency)) {
+      return pool.token0
```
hensha256, line 25: i think this reads weirdly because its not apparent that `currency0` and `token0` are the same thing

PR #147, opened 2024-10-03
feat(ur-sdk): Add V4 Parser to UR parser

marktoda:
This commit merges in the v4 parser as a subparser for UR calldata.

## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/universal-router-sdk/src/utils/routerCommands.ts
```diff
@@ -30,17 +30,35 @@ export enum CommandType {
...
 }
 
+export type CommandDefinition =
+  | {
+      parser: Parser.Abi
+      params: ParamType[]
+    }
+  | {
+      parser: Parser.V4Actions
+    }
```
marktoda, line 60: ok, I've done a bit of a refactor here because we previously were not describing what was actually happening. 

Universal Router command inputs are _arbitrary bytes_, that can mean anything but currently are either:
- ABI encoded parameters for some action
- direct calldata for a function on the V4 router or the V4 Position manager

Previously we were just describing the ABI, which was insufficient for actually expressing what was happening in the v4 case. This is why we had a hard-coded edge case in `createCommand`:
```
  if (type === CommandType.V4_SWAP) {
    return { type, encodedInput: parameters[0] }
  }
```

I've expanded the command inputs definition to capture this nuance, allowing different top level "parsers" for each command type's inputs

marktoda: cc @hensha256

hensha256: ohhh interesting interesting, have you merged main? diana made some changes to that logic just this afternoon so want to make sure thats taken into account. But from what I can see youve already included that

marktoda: yea I did! I made a separate `V3Actions` vs `V4Actions`, where both bypass the abi decoding

sdks/universal-router-sdk/test/utils/commandParser.test.ts
```diff
@@ -1,12 +1,40 @@
...
+    liquidityGross: amount,
+  },
+  {
+    index: nearestUsableTick(TickMath.MAX_TICK, 10),
+    liquidityNet: amount.mul(-1),
+    liquidityGross: amount,
+  },
+]
+const USDC = new Token(1, '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48', 6, 'USDC', 'USD Coin')
+const DAI = new Token(1, '0x6B175474E89094C44Da98b954EedeAC495271d0F', 18, 'DAI', 'DAI Stablecoin')
```
hensha256, line 26: unused?

hensha256 (approved):

sdks/universal-router-sdk/src/utils/commandParser.ts
```diff
@@ -127,3 +148,17 @@ export function parseV3PathExactOut(path: string): readonly V3PathItem[] {
 
   return res
 }
+
+function v4RouterCallToParams(actions: readonly V4RouterAction[]): readonly Param[] {
```
marktoda, line 152: open to feedback on this remapping of names. The alternative was to keep `command.params` as a V4RouterAction object like 

```
{
   actionName: SWAP_EXACT_IN,
   ActionType: V4Action.SWAP_EXACT_IN,
   params: [...[
}
```

opted for this to be more standardized with the other non-v4 command datas

hensha256 (approved):

PR #156, opened 2024-10-08
fix(universal-router-sdk): fix migration encoding

dianakocsis:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -106,14 +112,28 @@ export abstract class SwapRouter {
       invariant(universalRouterAddress == options.v3RemoveLiquidityOptions.permit.spender, 'INVALID_SPENDER')
       // don't need to transfer it because v3posm uses isApprovedOrOwner()
       encodeV3PositionPermit(planner, options.v3RemoveLiquidityOptions.permit, options.v3RemoveLiquidityOptions.tokenId)
+      // remove permit so that multicall doesnt add it again
+      delete options.v3RemoveLiquidityOptions.permit
```
marktoda, line 116: wait why don't you just remove it above?

marktoda (approved):

sdks/universal-router-sdk/src/swapRouter.ts
hensha256: because it needs to be encoded on line 114 above this as a separate UR command

Issue #141, opened 2024-10-03
Ethers V6 Upgrade


jfbloom22:
Any interest in updating the SDKs to Ethers V6?  I am really enjoying using BigInts rather than BigNumber.  Happy to help with this effort.  

Which SDKs need to be updated for Ethers V6? 
- [ ] permit2-sdk
- [ ] universal-router-sdk
- [ ] uniswapx-sdk
- [ ] v4-sdk

SergioArrighi:
+1

Surely v4-sdk.
I got there while trying to migrate ImmutableX ts sdk.

Regards

arcolife:
+1 on ethers v6 for uniswapx-sdk

Amxx:
I noticed that in some places, the SDK uses ethers v5 "directly" by importing `ethers/...` instead of `@ethersproject/...`
See: https://github.com/Uniswap/sdks/blob/a8713a0dcf59050670cecbecd77f4e2c5e5aebec/sdks/v4-sdk/src/entities/pool.ts#L13

This causes conflicts when using the SDK with other packages that depend on ethers v6.

Should I open a PR to fix these imports before the transition to v6 happens ?

haowang1013:
Can you guys at least declare the ethers.js v5 dependency correctly rather than assuming whatever version the depending package uses is correct? That way npm will install a v5 for your package even if the depending package uses v6.

andrewverify:
as per @haowang1013, need the uniswapX SDK to load v5 as a dependency in order to be able to use it with hardhat and other libraries that use v6. Right now we have to solve a lot of conflicts between the way v5 and v6 does things like BigNumber / BigInt conversion etc.

treeder:
I tried getting this working over the past couple of days and got it pretty close I think in this PR: https://github.com/treeder/uniswap-sdks/pull/1. 

But after spending way too much time and wrecking my keyboard with tears, I gave up. There's so much legacy stuff spread throughout, dead libraries that haven't been maintained in 5 years, etc. It's just a nightmare in general. If I had a week and didn't care about my time or sanity, I think it could be finished up. 

My new plan is to just use this sdk in a separate service and call it via an API as there's no way to get it to work with a modern stack.

jnsvd:
Also interested in a fix for this. Currently we can't require `@uniswap/smart-order-router` in projects where we have a hardhat installation that uses ethers v6.

yuercl:
Here is a method provided, but it is not an elegant solution.

[...]

PR #187, opened 2024-10-29
fix: avoid import conflicts between v4-sdk and ethers-v6

Amxx:
## Description

`v4-sdk` uses ethers v5, which is slowly going away. Many project that import `v4-sdk` may also have ethers-v6 in their dependency. This creates conflicts, because `v4-sdk` imports `ethers` directly (instead of importing `@ethersproject/xxx`).

## How Has This Been Tested?

This is only a change to the imports. There should be no feature changes. Was tested with `yarn test`

There shouldn't be any.

## Feedback Focus

The dependencies of the `v4-sdk` package do not include `ethers`. It only contains `@ethersproject/solidity`. However, `v4-sdk` does use features from all of the following
```
@ethersproject/abi
@ethersproject/address
@ethersproject/bignumber
@ethersproject/constants
@ethersproject/hash
@ethersproject/solidity
@ethersproject/units
@ethersproject/abstract-signer
```
Maybe they should be added to the dependencies in a more explicit way

## Follow Ups

Migration to ethers v6: https://github.com/Uniswap/sdks/issues/141

### DIFF
--- a/sdks/v4-sdk/src/utils/v4Planner.ts
+++ b/sdks/v4-sdk/src/utils/v4Planner.ts
@@ -1,6 +1,6 @@
 import invariant from 'tiny-invariant'
-import { defaultAbiCoder } from 'ethers/lib/utils'
-import { BigNumber } from 'ethers'
+import { defaultAbiCoder } from '@ethersproject/abi'
+import { BigNumber } from '@ethersproject/bignumber'
 import { Currency, Percent, TradeType } from '@uniswap/sdk-core'
 import { Trade } from '../entities/trade'
 import { ADDRESS_ZERO, EMPTY_BYTES } from '../internalConstants'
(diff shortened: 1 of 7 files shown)


sdks/v4-sdk/src/utils/v4Planner.ts
```diff
@@ -194,17 +194,17 @@ export class V4Planner {
     this.addAction(actionType, [
       exactOutput
         ? {
-            currencyOut,
-            path: encodeRouteToPath(trade.route, exactOutput),
-            amountInMaximum: trade.maximumAmountIn(slippageTolerance ?? new Percent(0)).quotient.toString(),
```
hensha256, line 199: youve changed some of the linting so CI is failing :)

jsy1218:
I will give a test to smart order router and routing-api on my local to make sure this PR is good

jsy1218:
I'm hitting weird issue on my local https://app.warp.dev/block/QhX8rzl8rm2lPoL2m7ysSQ. Although peeking this PR, I don't think it's causing the weird issue.

Amxx:
Vscode did some linting changes automatically. I just reverted them.

PR #197, opened 2024-11-13
feat(breaking): support Unichain Mainnet on v2 and v3

xrsv:
## PR Scope

Adds support for Unichain Mainnet enums/addresses on sdk-core
Introduces a breaking change for Unichain Sepolia enum name

## How Has This Been Tested?

Will test in routing/SOR

Yes - Introduces a breaking change for Unichain Sepolia enum name from `ASTROCHAIN_SEPOLIA` to `UNICHAIN_SEPOLIA`


sdks/sdk-core/src/addresses.ts
```diff
@@ -292,6 +294,16 @@ const ASTROCHAIN_SEPOLIA_ADDRESSES: ChainAddresses = {
   v4QuoterAddress: '0x2cfa87651D3AB05Bc59E325E5eaC8495CF34cE55',
 }
 
+const UNICHAIN_ADDRESSES: ChainAddresses = {
+  v3CoreFactoryAddress: '0x1f98400000000000000000000000000000000003',
+  multicallAddress: '0xb7610f9b733e7d45184be3a1bc966960ccc54f0b',
```
jsy1218, line 299: multical address is unichain sepolia address. looks like we haven't deployed [deploy-v3](https://github.com/Uniswap/deploy-v3)

sdks/sdk-core/src/chains.ts
```diff
@@ -25,6 +25,7 @@ export enum ChainId {
   ZKSYNC = 324,
   WORLDCHAIN = 480,
   ASTROCHAIN_SEPOLIA = 1301,
```
jsy1218, line 27: i've been thinking whether we should rename ASTROCHAIN_SEPOLIA to UNICHAIN_SEPOLIA, now we are here. Then make it a breaking change, by titling PR breaking(sdk-core): ...

sdks/sdk-core/src/addresses.ts
jsy1218: im wrong, they are from mainnet

sdks/sdk-core/src/chains.ts
dannydaniil: agreed!

dannydaniil (approved):
lgtm, assuming the addresses have been checked

sdks/sdk-core/src/chains.ts
xrsv: updated

cgkol (approved):

PR #206, opened 2024-11-21
feat(universal-router-sdk): support unichain mainnet on universal-router-sdk

xrsv:
## Description

Support unichain mainnet on universal router sdk

**Note**: UR address is temporary and will change again in near future

## How Has This Been Tested?

Will test in routing

No


cgkol (approved):

hensha256 (commented):
the address and creation block look correct to me, but is it correct that the same UR is used for `V1_2` and `V2_0`? I dont know what those specific versions mean in the sdk

xrsv:
> the address and creation block look correct to me, but is it correct that the same UR is used for `V1_2` and `V2_0`? I dont know what those specific versions mean in the sdk

@hensha256 yes this should be fine - see entries of other chains above, all have same address for those (except sepolia).
Those are going to come into play once v4 is introduced I believe ([related PR here ](https://github.com/Uniswap/sdks/pull/88)that introduce this split)

cgkol (approved):

jsy1218:
> the address and creation block look correct to me, but is it correct that the same UR is used for `V1_2` and `V2_0`? I dont know what those specific versions mean in the sdk

@hensha256 v1_2 means UR with v2 and v3, i.e., v2 everywhere. v2_0 means UR with v2 and v3 and v4. Unichain initial deploy will have a single UR with v2 and v3 and v4 right? Hence being the same seems no harm.

PR #210, opened 2024-11-28
fix(v4-sdk): remove take_settle_pair, use updated actions constants

snreynolds:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/v4-sdk/src/utils/v4Planner.ts
```diff
@@ -17,27 +17,26 @@ export enum Actions {
   DECREASE_LIQUIDITY = 0x01,
   MINT_POSITION = 0x02,
   BURN_POSITION = 0x03,
+
```
hensha256, line 20: i wonder if we should put in placeholder comments about why some numbers are missing?

hensha256 (approved):

sdks/v4-sdk/src/utils/v4Planner.ts
snreynolds: yeah will do!

sdks/v4-sdk/src/utils/abi.ts
```diff
@@ -3,70 +3,188 @@ export const abi = [
   {
     type: 'constructor',
     inputs: [
-      { name: '_poolManager', type: 'address', internalType: 'contract IPoolManager' },
-      { name: '_permit2', type: 'address', internalType: 'contract IAllowanceTransfer' },
-      { name: '_unsubscribeGasLimit', type: 'uint256', internalType: 'uint256' },
-      { name: '_tokenDescriptor', type: 'address', internalType: 'contract IPositionDescriptor' },
+      {
```
hensha256, line 6: can we rename this file to be clearer _what_ it is the ABI for? i had to read the ABI to figure it out lol

hensha256 (approved):

PR #211, opened 2024-12-02
fix(uniswapx-sdk): DutchV3 cosigner digest

codyborn:
## Description

Updating the contract bytecode and cosigner digest generation to the latest.

## How Has This Been Tested?

Local tests

No


ConjunctiveNormalForm (approved):
lgtm! Do we want to have a test here given that we are handrolling the fields here

alanhwu:
We have an integration test to test changes like this, it was what made us realize the abi needed updating

Fart10 (commented):
🙏

PR #212, opened 2024-12-05
feat(sdk-core): add monad testnet

BZahory:
## PR Scope

## Description

Adds Monad Testnet.

Unblocks review on https://github.com/Uniswap/universe/pull/14305

## How Has This Been Tested?

Linked to local repo and ensure everything runs as expected

No

## (Optional) Feedback Focus

## (Optional) Follow Ups

### DIFF
--- a/sdks/sdk-core/src/addresses.ts
+++ b/sdks/sdk-core/src/addresses.ts
@@ -303,7 +303,6 @@ const UNICHAIN_ADDRESSES: ChainAddresses = {
   tickLensAddress: '0xd5d76fa166ab8d8ad4c9f61aaa81457b66cbe443',
   swapRouter02Address: '0x73855d06de49d0fe4a9c42636ba96c62da12ff9c',
 }
-// exclude monad because it doesn't have v3 support
 export const CHAIN_TO_ADDRESSES_MAP: Record<SupportedChainsType, ChainAddresses> = {
   [ChainId.MAINNET]: MAINNET_ADDRESSES,
   [ChainId.OPTIMISM]: OPTIMISM_ADDRESSES,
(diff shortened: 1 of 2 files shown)


sdks/sdk-core/src/addresses.ts
```diff
@@ -303,8 +303,8 @@ const UNICHAIN_ADDRESSES: ChainAddresses = {
   tickLensAddress: '0xd5d76fa166ab8d8ad4c9f61aaa81457b66cbe443',
   swapRouter02Address: '0x73855d06de49d0fe4a9c42636ba96c62da12ff9c',
 }
-
-export const CHAIN_TO_ADDRESSES_MAP: Record<SupportedChainsType, ChainAddresses> = {
+// exclude monad because it doesn't have v3 support
+export const CHAIN_TO_ADDRESSES_MAP: Omit<Record<SupportedChainsType, ChainAddresses>, ChainId.MONAD_TESTNET> = {
```
jsy1218, line 307: change to `Omit<Record<SupportedChainsType, ChainAddresses>, ChainId.MONAD_TESTNET> ` doesn't seem like a great solution. So far we have relied on SupportedChainsType enum strong type to ensure we don't miss an address on any supported chain.

sdks/sdk-core/src/addresses.ts
```diff
@@ -303,8 +303,8 @@ const UNICHAIN_ADDRESSES: ChainAddresses = {
   tickLensAddress: '0xd5d76fa166ab8d8ad4c9f61aaa81457b66cbe443',
   swapRouter02Address: '0x73855d06de49d0fe4a9c42636ba96c62da12ff9c',
 }
-
-export const CHAIN_TO_ADDRESSES_MAP: Record<SupportedChainsType, ChainAddresses> = {
+// exclude monad because it doesn't have v3 support
```
jsy1218, line 306: this might be a blocker for routing. do we need why no v3 support? is it because of no v3 protocol deploy on monad testnet?

BZahory: they only want v2 for their competition: https://www.notion.so/uniswaplabs/Monad-Testnet-12fc52b2548b802b8012ff65a9b376ba?pvs=4#12fc52b2548b8071a11ae1c56ae73dcd

sdks/sdk-core/src/addresses.ts
BZahory: Removed this to unblock urgent wallet PR but we'll need to revisit (maybe launch v3 anyways?) for web

jsy1218 (approved):

sdks/sdk-core/src/addresses.ts
Fart10: 😎world

PR #276, opened 2025-01-28
feat(v4-sdk): additional options when migrating

dianakocsis:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups

### DIFF
--- a/sdks/v4-sdk/src/PositionManager.ts
+++ b/sdks/v4-sdk/src/PositionManager.ts
@@ -59,11 +59,6 @@ export interface MintSpecificOptions {
    * Initial price to set on the pool if creating
    */
   sqrtPriceX96?: BigintIsh
-
-  /**
-   * Whether the mint is part of a migration from V3 to V4.
-   */
-  migrate?: boolean
 }
 
 /**
@@ -130,6 +125,22 @@ export interface TransferOptions {
   tokenId: BigintIsh
 }
 
+export interface MigrateSpecificOptions {
+  /**
+   * Whether the mint is part of a migration from V3 to V4.
+   */
+  migrate: boolean;
+  /** 
+   * Whether the migrate needs additional transfer or not 
+  */
+  additionalTransfer?: AdditionalTransferDetails;
+}
+
+export interface AdditionalTransferDetails {
+  neededCurrency: Currency;
+  neededAmount: BigintIsh;
+}
+
 export interface PermitDetails {
   token: string
   amount: BigintIsh
@@ -182,9 +193,10 @@ export interface NFTPermitData {
 }
 
 export type MintOptions = CommonOptions & CommonAddLiquidityOptions & MintSpecificOptions
+export type MigrateOptions = MintOptions & MigrateSpecificOptions;
 export type IncreaseLiquidityOptions = CommonOptions & CommonAddLiquidityOptions & ModifyPositionSpecificOptions
 
-export type AddLiquidityOptions = MintOptions | IncreaseLiquidityOptions
+export type AddLiquidityOptions = MintOptions | IncreaseLiquidityOptions | MigrateOptions
 
 export type RemoveLiquidityOptions = CommonOptions & RemoveLiquiditySpecificOptions & ModifyPositionSpecificOptions
 
@@ -193,6 +205,10 @@ export type CollectOptions = CommonOptions & CollectSpecificOptions
 // type guard
 function isMint(options: AddLiquidityOptions): options is MintOptions {
   return Object.keys(options).some((k) => k === 'recipient')
+}
+
+function isMigrate(options: AddLiquidityOptions): options is MigrateOptions {
+  return Object.keys(options).some((k) => k === 'migrate')
 }
 
 function shouldCreatePool(options: MintOptions): boolean {
@@ -275,7 +291,7 @@ export abstract class V4PositionManager {
     }
 
     // If migrating, we need to settle and sweep both currencies individually
-    if (isMint(options) && options.migrate) {
+    if (isMigrate(options) && options.migrate) {
       // payer is v4 positiion manager
       planner.addSettle(position.pool.currency0, false)
       planner.addSettle(position.pool.currency1, false)


sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -59,11 +59,6 @@ export interface MintSpecificOptions {
    * Initial price to set on the pool if creating
    */
   sqrtPriceX96?: BigintIsh
-
-  /**
-   * Whether the mint is part of a migration from V3 to V4.
-   */
-  migrate?: boolean
```
Ayoakala, line 66: looks like we are pulling out migrate into its own object; might need to change trading api impl

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -130,6 +125,23 @@ export interface TransferOptions {
...
+  /**
+   * Whether the migrate needs additional transfer or not
+   */
+  additionalTransfer?: AdditionalTransferDetails
+}
+
+export interface AdditionalTransferDetails {
+  neededCurrency: Currency
+  neededAmount: BigintIsh
+}
```
Ayoakala, line 143: not too bad it seems on the Trading API side to send in; curious how to derive the amount, seems it wouldnt bee to tedious to compute from earlier convos

Ayoakala (commented):
LGTM; so far will let someone on protocols do the approval honors

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -62,9 +62,9 @@ export interface MintSpecificOptions {
   sqrtPriceX96?: BigintIsh
 
   /**
-   * Whether the mint is part of a migration from V3 to V4.
+   * Whether the mint is part of a migration from V3 to V4 and the additional currency and amount to send if needed
    */
-  migrate?: boolean
+  migrateOptions?: MigrateOptions
```
dianakocsis, line 67: was thinking if we like this format, we could add the recipient field here too

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -114,21 +114,19 @@ export interface CollectSpecificOptions {
...
   /**
-   * The account sending the NFT.
+   * Whether the mint is part of a migration from V3 to V4.
    */
-  sender: string
-
+  migrate?: boolean
   /**
-   * The account that should receive the NFT.
+   * The additional currency that needs to be transferred if migrating from out of range to in range or out of range to opposite side out of range
```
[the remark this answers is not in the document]
hensha256: something like this? i found the original sentence hard to read with no punctuation

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -114,21 +114,19 @@ export interface CollectSpecificOptions {
...
    */
-  recipient: string
-
+  neededCurrency?: Currency
```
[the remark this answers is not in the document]
hensha256: and same with neededAmount

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -284,9 +282,11 @@ export abstract class V4PositionManager {
 
     let value: string = toHex(0)
 
+    let needToSendEth = isMint(options) && options.migrateOptions?.neededCurrency?.isNative
+
     // If migrating, we need to settle and sweep both currencies individually
-    if (isMint(options) && options.migrate) {
-      if (options.useNative) {
+    if (isMint(options) && options.migrateOptions?.migrate) {
+      if (options.useNative && !needToSendEth) {
```
hensha256, line 289: in the case where neededCurrency is ETH and the position is in WETH this path should still be taken i think?

sdks/v4-sdk/src/PositionManager.test.ts
```diff
@@ -367,6 +367,45 @@ describe('PositionManager', () => {
       expect(value).toEqual('0x00')
     })
 
+    it('succeeds when migrating out of range and need to transfer eth', () => {
```
hensha256, line 370: would love a test for non-eth too, and for ETH-WETH wrapping combos if they are to be supported too

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -284,9 +282,11 @@ export abstract class V4PositionManager {
 
     let value: string = toHex(0)
 
+    let needToSendEth = isMint(options) && options.migrateOptions?.neededCurrency?.isNative
```
hensha256, line 285: its weird to me that `migrateOptions.migrate` could be `false`, but then `neededCurrency` could still be set?
By definition if `migrateOptions.neededCurrency` is set, you're migrating no? It feels like either `migrate` should be a boolean OUTSIDE of the options, and then if `migrate`, you check `options`.
OR you should have an invariant that says if `neededCurrency` is set, `migrate` must be true

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -114,21 +114,19 @@ export interface CollectSpecificOptions {
...
    */
-  recipient: string
-
+  neededCurrency?: Currency
   /**
-   * The id of the token being sent.
+   * The amount of additional currency that needs to be transferred if migrating from out of range to in range or out of range to opposite side out of range
    */
-  tokenId: BigintIsh
+  neededAmount?: BigintIsh
```
hensha256, line 129: Should `neededAmount` and `neededCurrency` be combined into a single `CurrencyAmount` type?
You should never have one without the other, so i think a single optional `CurrencyAmount` would make more sense

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -114,21 +114,19 @@ export interface CollectSpecificOptions {
   recipient: string
 }
 
-export interface TransferOptions {
```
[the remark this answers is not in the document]
dianakocsis: there was a SafeTransferOptions in v3, doesn't seem like TransferOptions was used for anything in v4 tho

sdks/v4-sdk/src/PositionManager.ts
dianakocsis: I think the opposite right? it would need to wrap not unwrap

sdks/v4-sdk/src/PositionManager.ts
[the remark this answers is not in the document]
dianakocsis: For 2, you mean unwrapping right?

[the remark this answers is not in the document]
dianakocsis: i was thinking the trading api would calculate how much is needed of the currency in the new position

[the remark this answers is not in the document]
dianakocsis: but i guess the sdk could handle all scenarios?

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -114,21 +114,22 @@ export interface CollectSpecificOptions {
...
+}
 
+export interface CurrencyAmount {
   /**
-   * The account that should receive the NFT.
+   * The additional currency that needs to be transferred
    */
-  recipient: string
-
+  inputCurrency: Currency
```
snreynolds, line 128: Can you explain why there is a need for a 3rd currency specification? If we're migrating from v3 we can only ever migrate to a pool that contains the same currencies no?

And if we detect the pool is weth, we can just auto-unwrap to eth?

PR #277, opened 2025-01-28
feat(ur-sdk): migrating out of range

dianakocsis:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -151,16 +145,31 @@ export abstract class SwapRouter {
       planner.addCommand(CommandType.V3_POSITION_MANAGER_CALL, [v3Call])
     }
 
+    // if migrate options has a currency, require a batch permit
```
Ayoakala, line 148: nit: if migrate options has an additionalTransfer oprion, require a batch permit

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -151,16 +145,31 @@ export abstract class SwapRouter {
...
+        options.migrateOptions.batchPermit.permitBatch,
+        options.migrateOptions.batchPermit.signature,
+      ])
+      planner.addCommand(CommandType.PERMIT2_TRANSFER_FROM, [
+        options.migrateOptions.additionalTransfer.neededCurrency,
+        options.v3RemoveLiquidityOptions.collectOptions.recipient,
+        options.migrateOptions.additionalTransfer.neededAmount,
+      ])
+      delete options.migrateOptions.batchPermit
+    }
```
Ayoakala, line 161: what would this batch permit look like; I already have a [function](https://github.com/Uniswap/external-api/blob/206d6bed13218df635cfb66583b9e01a0ba5deff/lib/calculators/TokenPermitCalculator.ts#L71-L106) for creating v4 batch permits; Im assuming that this will be the same one needed here since it is also permitting to add funds. Although the spender here seems to be UR instead of the Position manager

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
```diff
@@ -89,7 +91,7 @@ contract MigratorCallParametersTest is Test, Interop, DeployRouter {
         // add the position to v3 so we have something to migrate
         assertEq(INonfungiblePositionManager(V3_POSITION_MANAGER).balanceOf(from), 0);
         // USDC < WETH
```
Ayoakala, line 93: in range as well?

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
```diff
@@ -107,4 +109,204 @@ contract MigratorCallParametersTest is Test, Interop, DeployRouter {
...
+        mintV3Position(address(USDC), address(WETH), 3000, 2500e6, 1e18, 205320, 300000);
+        assertEq(INonfungiblePositionManager(V3_POSITION_MANAGER).balanceOf(from), 1);
+
+        // approve the UniversalRouter to access the position (instead of permit)
+        vm.startPrank(from);
+        INonfungiblePositionManager(V3_POSITION_MANAGER).setApprovalForAll(MAINNET_ROUTER, true);
+        vm.stopPrank();
+
+        assertEq(params.value, 0);
+        WETH.transfer(from, WETH.balanceOf(address(this)));
```
Ayoakala, line 129: Im guessing this is the second permit / `additionalTransfer` batch permit that will permit the user to send in more funds right?

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
```diff
@@ -107,4 +109,204 @@ contract MigratorCallParametersTest is Test, Interop, DeployRouter {
...
+        // add the position to v3 so we have something to migrate
+        assertEq(INonfungiblePositionManager(V3_POSITION_MANAGER).balanceOf(from), 0);
+        // USDC < WETH
+        // one sided v3 position in USDC
+        mintV3Position(address(USDC), address(WETH), 3000, 2500e6, 1e18, 205320, 300000);
+        assertEq(INonfungiblePositionManager(V3_POSITION_MANAGER).balanceOf(from), 1);
+
+        // approve the UniversalRouter to access the position (instead of permit)
+        vm.startPrank(from);
+        INonfungiblePositionManager(V3_POSITION_MANAGER).setApprovalForAll(MAINNET_ROUTER, true);
```
Ayoakala, line 195: no extra transfer permit here because they are both on the same side/range right?

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
```diff
@@ -107,4 +109,204 @@ contract MigratorCallParametersTest is Test, Interop, DeployRouter {
...
+        vm.stopPrank();
+
+        assertEq(params.value, 0);
+        WETH.transfer(from, WETH.balanceOf(address(this)));
+        // approve permit2 to spend WETH
+        vm.startPrank(from);
```
Ayoakala, line 131: nit: might be wrong here but should we be standardizing how we call `prank`? I see we use `vm.prank(from)` in some places and `vm.startPrank(from)` + `vm.stopPrank()` in others; curious the difference between them

Ayoakala (commented):
LGTM; will let someone on protocols do final approvals

Had that question on the new batch permit and wanted to verify

sdks/universal-router-sdk/src/swapRouter.ts
dianakocsis: nice

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
dianakocsis: this is actually in range -> out of range. will comment!

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
dianakocsis: - `WETH.transfer(from, WETH.balanceOf(address(this)));` is so the user has WETH to begin with in order to send in
- `WETH.approve(MAINNET_PERMIT2, WETH.balanceOf(from));` is because the user needs to approve the PERMIT2 contract first to be able to use it. is that okay?

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
dianakocsis: yes exactly

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
dianakocsis: vm.prank() is just for one call, while vm.startPrank() is until you call vm.stopPrank() - let me see if i can make it look cleaner anywhere!

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -151,16 +145,31 @@ export abstract class SwapRouter {
       planner.addCommand(CommandType.V3_POSITION_MANAGER_CALL, [v3Call])
```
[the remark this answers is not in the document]
dianakocsis: no risk! we're just deleting createPool, permit, and batchPermit since they are options that could get encoded again later

sdks/universal-router-sdk/src/swapRouter.ts
dianakocsis: yes exactly! does that work?

dianakocsis: this also requires that the user has already approved the permit2 contract to spend their currency. is this checked anywhere?
also im thinking sending the batch permit might not be required if they have permitted the universal router over their currency in a different transaction. is there a way to check this?

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -162,6 +162,28 @@ export abstract class SwapRouter {
...
 
+    // if migrate options has a currency, require a batch permit
+    if (options.v4AddLiquidityOptions.migrateOptions?.neededCurrency) {
+      // need to permit the UR to spend your currency on permit2
+      if (options.v4AddLiquidityOptions.batchPermit) {
+        invariant(
+          options.v4AddLiquidityOptions.batchPermit.permitBatch.spender == universalRouterAddress,
+          'INVALID_SPENDER'
+        )
+        planner.addCommand(CommandType.PERMIT2_PERMIT_BATCH, [
```
hensha256, line 173: its only ever going to be 1 currency i think? so using `batch` is unnecessary?

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
```diff
@@ -318,4 +323,332 @@ contract MigratorCallParametersTest is Test, Interop, DeployRouter {
...
+        assertEq(address(RECIPIENT).balance, recipientBalanceBefore);
+        assertGe(USDC.balanceOf(RECIPIENT), recipientUSDCBalanceBefore);
+        assertGt(WETH.balanceOf(RECIPIENT), recipientWETHBalanceBefore);
+
+        // old position burned, new position minted
+        assertEq(INonfungiblePositionManager(V3_POSITION_MANAGER).balanceOf(from), 0, "V3 NOT BURNT");
+        assertEq(v4PositionManager.balanceOf(RECIPIENT), 1, "V4 NOT MINTED");
+    }
+
+    function test_migrate_v3OutOfRangeIn1_v4OutOfRangeIn1() public {
```
hensha256, line 603: do we need tests for unwrapping/wrapping at all?

sdks/universal-router-sdk/src/swapRouter.ts
dianakocsis: permitBatch is already a param of v4AddLiquidityOptions so i figured i would just keep it instead of adding another one?

PR #278, opened 2025-01-28
feat(v4-sdk): migrate weth to eth by unwrapping

dianakocsis:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


Ayoakala (commented):
LGTM; should we add a test?

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -276,25 +280,30 @@ export abstract class V4PositionManager {
 
     // If migrating, we need to settle and sweep both currencies individually
     if (isMint(options) && options.migrate) {
-      // payer is v4 positiion manager
+      if (options.useNative) {
+        planner.addUnwrap(BigNumber.from('0x8000000000000000000000000000000000000000000000000000000000000000'))
```
snreynolds, line 284: this should be defined as a constant

snreynolds: also its not always the case you will want to unwrap the full amount no? maybe it complicates things but by unwrapping the full amount refunds from a weth pool will be in eth

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -276,25 +280,30 @@ export abstract class V4PositionManager {
...
+      }
+      // payer is v4 position manager
       planner.addSettle(position.pool.currency0, false)
       planner.addSettle(position.pool.currency1, false)
       planner.addSweep(position.pool.currency0, options.recipient)
       planner.addSweep(position.pool.currency1, options.recipient)
+      // sweep wrapped native for good measure
+      if (options.useNative) {
+        planner.addSweep(position.pool.currency0.wrapped, options.recipient)
```
snreynolds, line 293: this doesn't make sense if we unwrap the full contract balance

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -276,25 +280,30 @@ export abstract class V4PositionManager {
...
     let value: string = toHex(0)
-    if (options.useNative) {
-      invariant(position.pool.currency0.isNative || position.pool.currency1.isNative, NO_NATIVE)
-      let nativeCurrency: Currency = position.pool.currency0.isNative
-        ? position.pool.currency0
-        : position.pool.currency1
-      value = position.pool.currency0.isNative ? toHex(amount0Max) : toHex(amount1Max)
-      planner.addSweep(nativeCurrency, MSG_SENDER)
+    // When not migrating and adding native currency, add a final sweep
+    if (options.useNative && !(isMint(options) && options.migrate)) {
```
snreynolds, line 303: I'm confused on this case? why do we need to add  `!(isMint(options) && options.migrate)`

and why do we take out the invariants?

snreynolds: Oh i think we check it above?

snreynolds: eek just finding the if else branches hard to reason though gonna think if theres a way to clean it up a tad

dianakocsis: i took out `|| position.pool.currency1.isNative` because this will never be the case - if one of the currencies is native, it will always be currency0

dianakocsis: added this: `!(isMint(options) && options.migrate)` because we should only add an extra sweep if we are not migrating. The 2 sweeps are already covered when migrating, so we don't need to add an extra

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -240,6 +241,9 @@ export abstract class V4PositionManager {
       calldataList.push(V4PositionManager.encodeInitializePool(position.pool.poolKey, options.sqrtPriceX96!))
     }
 
+    // undefined is treated as false for useNative
+    invariant(position.pool.currency0.isNative === !!options.useNative, NO_NATIVE)
```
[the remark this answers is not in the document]
dianakocsis: okay but this would only be true if adding eth. the first option ensures that currency0 is native if and only if the useNative flag is set

sdks/v4-sdk/src/PositionManager.ts
dianakocsis: could change to `planner.addUnwrap(BigNumber.from(0))` and keep the weth sweep at the end

dianakocsis: that way it gets just the contract's debt instead of the whole balance, and refunds could be kept in weth

dianakocsis: the trading api calculates the liquidity to add based off of the 2 amounts received from v3, so im not sure when the contract balance wouldnt equal the correct amount tho. but could do this for good measure

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -73,7 +74,7 @@ export interface CommonAddLiquidityOptions {
   /**
    * Whether to spend ether. If true, one of the currencies must be the NATIVE currency.
    */
-  useNative?: NativeCurrency
+  useNative?: boolean
```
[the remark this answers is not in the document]
hensha256: its definitely more understandable as a boolean imo `useNative` is a boolean name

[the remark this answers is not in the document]
hensha256: how come this option existed before if it didnt support using native?

sdks/v4-sdk/src/PositionManager.ts
hensha256: yeah i think it makes sense to do it with OPEN_DELTA instead

sdks/v4-sdk/src/PositionManager.ts
hensha256: why cant it just be `&& !options.migrate`? Why do you need to check the mint options too?

dianakocsis: typescript is annoying and says that migrate doesnt exist on addLiquidityOptions since it doesnt exist on IncreaseLiquidityOptions. so u need to make sure its a mint

sdks/v4-sdk/src/PositionManager.ts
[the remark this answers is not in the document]
snreynolds: i lean towards keeping it though as close to the v3-sdk as possible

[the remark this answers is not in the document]
snreynolds: this changes that interface, we should just leave it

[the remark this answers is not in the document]
dianakocsis: yeah it's so much more understandable as a boolean, but i do agree it would be nice to keep it like v3-sdk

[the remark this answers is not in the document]
dianakocsis: let me see how i can rearrange

sdks/v4-sdk/src/PositionManager.ts
snreynolds: do we not allow partial liquidity deposits?

snreynolds: even if not there could 100% be a case where because of price differences the amount of weth in the v3 position > the amount of eth required in the v4 position

dianakocsis: when migrating u need to migrate 100%

sdks/v4-sdk/src/PositionManager.ts
snreynolds: this makes sense now!! cool

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -276,25 +284,30 @@ export abstract class V4PositionManager {
...
     let value: string = toHex(0)
-    if (options.useNative) {
-      invariant(position.pool.currency0.isNative || position.pool.currency1.isNative, NO_NATIVE)
-      let nativeCurrency: Currency = position.pool.currency0.isNative
-        ? position.pool.currency0
-        : position.pool.currency1
-      value = position.pool.currency0.isNative ? toHex(amount0Max) : toHex(amount1Max)
-      planner.addSweep(nativeCurrency, MSG_SENDER)
+    // When not migrating and adding native currency, add a final sweep
+    if (options.useNative && !(isMint(options) && options.migrate)) {
```
snreynolds, line 307: just dpuble triple confirming, in the migrate eth case we do not need to sweep eth since we are only unwrapping and sending in the entire contract balance so there is never any eth leftover... any edge cases here?

dianakocsis: okay ya ur right we just got some gnarly if else statements going on

dianakocsis: because now we need another if else statement at the end on whether to sweep currency0 or currency0.wrapped

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -276,25 +284,34 @@ export abstract class V4PositionManager {
...
+      } else {
+        // payer is v4 position manager
+        planner.addSettle(position.pool.currency0, false)
+        planner.addSettle(position.pool.currency1, false)
+        planner.addSweep(position.pool.currency0, options.recipient)
+        planner.addSweep(position.pool.currency1, options.recipient)
+      }
     } else {
       // need to settle both currencies when minting / adding liquidity (user is the payer)
       planner.addSettlePair(position.pool.currency0, position.pool.currency1)
```
snreynolds, line 305: think you can just add this if case here and delete it from below


```suggestion
      planner.addSettlePair(position.pool.currency0, position.pool.currency1)
      // Any sweeping must happen after the settling.
      let value: string = toHex(0)
     // When not migrating and adding native currency, add a final sweep
     if (options.useNative && !(isMint(options) && options.migrate)) {
     value = toHex(amount0Max)
     // native currency will always be currency0 in v4
     planner.addSweep(position.pool.currency0, MSG_SENDER)
    }
```

dianakocsis: nice!

snreynolds (approved):

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -11,12 +11,13 @@ import invariant from 'tiny-invariant'
...
   EMPTY_BYTES,
   CANNOT_BURN,
-  NO_NATIVE,
+  NATIVE_NOT_SET,
   NO_SQRT_PRICE,
   ONE,
   PositionFunctions,
   ZERO,
   ZERO_LIQUIDITY,
+  OPEN_DELTA,
```
hensha256, line 20: nit: this should be inserted in alphabetical order

sdks/v4-sdk/src/PositionManager.test.ts
```diff
@@ -312,6 +326,46 @@ describe('PositionManager', () => {
...
+        1,
+        toHex(amount0Max),
+        toHex(amount1Max),
+        recipient,
+        EMPTY_BYTES,
+      ])
+
+      planner.addAction(Actions.UNWRAP, [0])
+      planner.addAction(Actions.SETTLE, [toAddress(pool_1_eth.currency0), 0, false])
+      planner.addAction(Actions.SETTLE, [toAddress(pool_1_eth.currency1), 0, false])
```
hensha256, line 361: nit: would be more readable to use OPEN_DELTA in these 3 commands

sdks/v4-sdk/src/PositionManager.ts
```diff
@@ -274,27 +282,36 @@ export abstract class V4PositionManager {
...
+        // sweep any leftover wrapped native that was not unwrapped
+        planner.addSweep(position.pool.currency0.wrapped, options.recipient)
+        planner.addSweep(position.pool.currency1, options.recipient)
+      } else {
+        // payer is v4 position manager
+        planner.addSettle(position.pool.currency0, false)
+        planner.addSettle(position.pool.currency1, false)
+        planner.addSweep(position.pool.currency0, options.recipient)
+        planner.addSweep(position.pool.currency1, options.recipient)
+      }
```
hensha256, line 304: these `if` and `else` blocks are so similar and most of the lines are the same, i'd rather it wasn't duplicating so much logic 

```suggestion
        // unwrap the exact amount needed to send to the pool manager
        if (options.useNative) planner.addUnwrap(OPEN_DELTA)

        // payer is v4 position manager
        planner.addSettle(position.pool.currency0, false)
        planner.addSettle(position.pool.currency1, false)
        // for native sweep leftover wrapped native that wasnt unwrapped
        planner.addSweep(options.useNative ? position.pool.currency0.wrapped : position.pool.currency0, options.recipient)
        planner.addSweep(position.pool.currency1, options.recipient)
```

hensha256 (approved):

PR #279, opened 2025-01-28
feat(ur-sdk): allow migrating to eth position

dianakocsis:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -82,14 +83,26 @@ export abstract class SwapRouter {
...
-    invariant(token0 === v4Pool.token0, 'TOKEN0_MISMATCH')
-    invariant(token1 === v4Pool.token1, 'TOKEN1_MISMATCH')
+    // cannot migrate into a WETH position
+    if (token0 === WETH || token1 === WETH) {
+      invariant(v4Pool.token0.isNative, 'ETH_REQUIRED_TO_ADD')
+      invariant(options.v4AddLiquidityOptions.useNative, 'USE_NATIVE_FLAG_REQUIRED')
+      if (token0 === WETH) {
+        invariant(token1 == v4Pool.token1, 'TOKEN_MISMATCH')
+      } else {
+        invariant(token0 == v4Pool.token1, 'TOKEN_MISMATCH')
```
Ayoakala, line 98: Should this be?
```
token0 == v4Pool.token0
```

Or should the above be updated?

Might be good to add a comment here to prevent confusion

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -82,14 +83,26 @@ export abstract class SwapRouter {
...
+      }
+    } else {
+      // validate the parameters
+      invariant(token0 === v4Pool.token0, 'TOKEN0_MISMATCH')
+      invariant(token1 === v4Pool.token1, 'TOKEN1_MISMATCH')
+    }
+
```
Ayoakala, line 105: Might want to turn this into a helper method for readability + add tests for this to make sure we are doing the right validations for each case

Ayoakala (commented):
Left a couple comments for token mismatch invariants and validation

sdks/universal-router-sdk/src/swapRouter.ts
Ayoakala: and maybe we can simplify these predicates?

dianakocsis: no because ETH is always token0 in v4!

sdks/universal-router-sdk/src/swapRouter.ts
dianakocsis: yess will def add helper and or simplify

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -82,14 +83,30 @@ export abstract class SwapRouter {
...
+    // if migrating wrapped native -> native
+    if (nativeCurrency.wrapped.equals(token0) || nativeCurrency.wrapped.equals(token1)) {
+      // v4 pool must use native currency and must have useNative flag set
+      invariant(v4Pool.token0.isNative === options.v4AddLiquidityOptions.useNative, 'NATIVE_REQUIRED')
+      // set token0 or token1 to native currency to validate parameters later
+      if (nativeCurrency.wrapped.equals(token0)) {
+        token0 = nativeCurrency
+      } else if (nativeCurrency.wrapped.equals(token1)) {
+        token1 = token0
+        token0 = nativeCurrency
```
snreynolds, line 102: this seems unnecessary? we already validate that the pool must be a native pool, which means token0 will already be the nativeCurrency

snreynolds: oh I see its just to make the invariant pass since its checking the INPUT tokens. we should just write a different invariant. In this case its not always true that input.token0 =/= output.token0 || input.token1 =/output.token1, 

so we should just change the invariant to handle this case!

it should be something like 
wrappedToken = input.token0 == native.wrapped ? input.token0 : input.token1
invariant(wrappedToken == native.wrapped && input.nonWrappedToken == output.token1)

hensha256: isnt token0 always native and never token1? because native==address(0)?

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -82,14 +83,30 @@ export abstract class SwapRouter {
...
-    const token1 = options.inputPosition.pool.token1
+    let token0: Currency = options.inputPosition.pool.token0
+    let token1: Currency = options.inputPosition.pool.token1
     const v4PositionManagerAddress =
       positionManagerOverride ?? CHAIN_TO_ADDRESSES_MAP[v4Pool.chainId as SupportedChainsType].v4PositionManagerAddress
 
+    const nativeCurrency = Ether.onChain(v4Pool.chainId)
+
+    // if migrating wrapped native -> native
+    if (nativeCurrency.wrapped.equals(token0) || nativeCurrency.wrapped.equals(token1)) {
```
hensha256, line 94: this is hard to read because its unclear that `token0` is of the v3 pool not v4. I think some renaming could help make that clearer that these tokens are v3

hensha256: also just because a v3 pool has WETH, doesnt mean that the user wants to migrate to ETH? this fi statement will enter for WETH -> WETH migrations too wont it?

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -82,14 +83,30 @@ export abstract class SwapRouter {
...
+      // v4 pool must use native currency and must have useNative flag set
+      invariant(v4Pool.token0.isNative === options.v4AddLiquidityOptions.useNative, 'NATIVE_REQUIRED')
```
hensha256, line 96: the comment above this doesnt align with this check? shouldnt it use `&` not `===`?

sdks/universal-router-sdk/src/swapRouter.ts
dianakocsis: but in v3, weth could be token0 or token1. so if weth is token1, then we need to make sure that token0 in v3 = token1 in v4

hensha256: righttt my confusion comes from [this naming](https://github.com/Uniswap/sdks/pull/279#discussion_r1939918631), proof we need to rename i guess 😆

sdks/universal-router-sdk/src/swapRouter.ts
dianakocsis: ah okay i just asked if we are strictly preventing this case or not

snreynolds (approved):
this change seems fine but I think the title of this PR is now misleading? we just now allow migrating weth -> eth but we don't specifically disallow weth positions on v4?

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -82,14 +82,25 @@ export abstract class SwapRouter {
...
+    const v3Token1 = options.inputPosition.pool.token1
     const v4PositionManagerAddress =
       positionManagerOverride ?? CHAIN_TO_ADDRESSES_MAP[v4Pool.chainId as SupportedChainsType].v4PositionManagerAddress
 
+    const nativeCurrency = Ether.onChain(v4Pool.chainId)
+
     // validate the parameters
-    invariant(token0 === v4Pool.token0, 'TOKEN0_MISMATCH')
-    invariant(token1 === v4Pool.token1, 'TOKEN1_MISMATCH')
+    if (nativeCurrency.equals(v4Pool.currency0)) {
```
hensha256, line 93: is nativeCurrency needed for this? could it just be `v4Pool.currency0.isNative`?

hensha256: then you could remove the line above and the import too

sdks/universal-router-sdk/test/uniswapTrades.test.ts
```diff
@@ -1604,7 +1617,7 @@ describe('Uniswap', () => {
           tickUpper: 300000,
         }),
         outputPosition: new V4Position({
-          pool: WETH_USDC_V4,
+          pool: ETH_USDC_V4,
```
hensha256, line 1620: all of the successful migration tests are now WETH->ETH tests, theres no longer any tests of ERC20->ERC20 regular migration case

sdks/universal-router-sdk/test/forge/MigratorCallParameters.t.sol
```diff
@@ -85,8 +114,33 @@ contract MigratorCallParametersTest is Test, Interop, DeployRouter {
...
+
+        // all funds were swept out of contracts
+        assertEq(USDC.balanceOf(MAINNET_ROUTER), 0);
+        assertEq(WETH.balanceOf(MAINNET_ROUTER), 0);
+        assertEq(USDC.balanceOf(address(v4PositionManager)), 0);
+        assertEq(WETH.balanceOf(address(v4PositionManager)), 0);
+
+        // old position burned, new position minted
+        assertEq(INonfungiblePositionManager(V3_POSITION_MANAGER).balanceOf(from), 0, "V3 NOT BURNT");
+        assertEq(v4PositionManager.balanceOf(RECIPIENT), 1, "V4 NOT MINTED");
```
hensha256, line 139: can you check the ETH balance of v4 didnt change, and the WETH balance did? to show it migrated to WETH not ETH?

snreynolds (approved):

snreynolds (approved):

PR #280, opened 2025-01-29
feat(uniswapx-sdk): Add WETH map by chain

codyborn:
## PR Scope

Add weth chainId to address map.


codyborn:
Using this instead:https://github.com/Uniswap/sdks/blob/ccc05875a326a813afbabe940bf58751feb4fac5/sdks/sdk-core/src/entities/weth9.ts

PR #284, opened 2025-01-31
feat(router-sdk): add getter for amounts which returns native amounts in a trade

zhongeric:
## PR Scope

## Description

- Router Trade object now supports multiple swaps where the `inputAmount` != `pathInput`, or the first currency in a swap / route
- Also exposes a new getter `amounts` which returns the sum of native amounts in input/output

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/router-sdk/src/entities/trade.ts
```diff
@@ -155,13 +155,49 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
...
 
     this._outputAmount = totalOutputFromRoutes
     return this._outputAmount
   }
 
+  /**
+   * Returns the sum of all swaps within the trade where the pathInput/Output is native
+   * @returns
+   * inputAmount: total input amount
+   * inputAmountNative: total amount of native currency required for ETH input paths
```
snreynolds, line 169: so these should only ever be nonzero when the route is v4, yes?

snreynolds: or sorry i guess im confusing this with pathInput/pathOutput... 

these actually return the route input/output amounts so native could be non zero for v2/v3, and then require a wrap/unwrap.. yes?

sdks/router-sdk/src/entities/trade.test.ts
```diff
@@ -210,6 +246,10 @@ describe('Trade', () => {
       const trade = await Trade.fromRoute(route, amount, TradeType.EXACT_INPUT)
       expect(trade.inputAmount.currency).toEqual(ETHER)
       expect(trade.outputAmount.currency).toEqual(token0)
+
+      expect(trade.amounts.inputAmountNative).toBeDefined()
```
snreynolds, line 250: could assert that its 10?

zhongeric: no because the inputAmountNative reflects the ETH that must be used for the route, and since this is a WETH input route its 0

sdks/router-sdk/src/entities/trade.ts
zhongeric: native should always be 0 for v2/v3

sdks/router-sdk/src/entities/trade.test.ts
snreynolds: is the following true:
- if no weth input or output then inputAmount/outputAmount are undefined.
- if weth input, inputAmountNative is 0 and inputAmount is nonzero & defined
- if weth output, outputAmountNative is 0 and outputAmount is nonzero & defined
- if eth input, inputAmount nonzero & defined ? (only can be true on v4)
- if eth output, outputAmount nonzero & defined ? (only can be true on v4)
- if the route has both eth/weth in the route, then it must be entirely on v4 or a mixed route that includes a v4 path and inputAmount, outputAmount, inputAmountNative, and ouputAmountNative are all defined and non zero

snreynolds: confirming that the following is true (and when we say input/output here I mean path input/output):

- if no weth input or output then inputAmountNative/outputAmountNative are undefined.
- if weth input, inputAmountNative is 0 and inputAmount is nonzero & defined (no change from legacy)
- if weth output, outputAmountNative is 0 and outputAmount is nonzero & defined (no change from legacy)
- if eth input, inputAmountNative nonzero & defined ? (only can be true on v4)
- if eth output, outputAmountNative nonzero & defined ? (only can be true on v4)
- if the route has both eth/weth in the route, then it must be entirely on v4 or a mixed route that includes a v4 path and inputAmount, outputAmount, inputAmountNative, and ouputAmountNative are all defined and non zero

snreynolds (approved):

sdks/router-sdk/src/entities/trade.test.ts
hensha256: > if weth input, inputAmountNative is 0 and inputAmount is nonzero & defined (no change from legacy)

in this case if the true input (not path input) is WETH too, isnt `inputAmountNative` is undefined?

snreynolds: yes undefined in that case

sdks/router-sdk/src/entities/route.ts
```diff
@@ -7,6 +7,20 @@ import { Protocol } from './protocol'
 import { Currency, Price, Token } from '@uniswap/sdk-core'
 import { MixedRouteSDK } from './mixedRoute/route'
 
+// Helper function to get the pathInput and pathOutput for a V2 / V3 route
+// Copied from V4-sdk, but modified to work with V2 / V3 as they only support tokens not NativeCurrency
+export function getPathToken(currency: Currency, pool: Pair | V3Pool): Token {
+  if (pool.involvesToken(currency.wrapped)) {
+    return currency.wrapped
```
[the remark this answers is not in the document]
zhongeric: ah yeah I think you're right, I updated it to check against .wrapped of each token

sdks/router-sdk/src/entities/trade.ts
```diff
@@ -139,10 +139,10 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
       return this._inputAmount
     }
 
-    const inputCurrency = this.swaps[0].inputAmount.currency
+    const inputAmountCurrency = this.swaps[0].inputAmount.currency
     const totalInputFromRoutes = this.swaps
-      .map(({ inputAmount }) => inputAmount)
-      .reduce((total, cur) => total.add(cur), CurrencyAmount.fromRawAmount(inputCurrency, 0))
+      .map(({ inputAmount: routeInputAmount }) => routeInputAmount)
+      .reduce((total, cur) => total.add(cur), CurrencyAmount.fromRawAmount(inputAmountCurrency, 0))
```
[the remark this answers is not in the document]
zhongeric: no this is the original logic, the case we're worried about is ETH input, WETH pathInput:
- route.input is denoted in ETH
- swap.inputAmount is also denoted in ETH
despite the route starting in WETH

so this will sum up all inputs which are all ETH

sdks/router-sdk/src/entities/trade.ts
```diff
@@ -155,13 +155,49 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
...
+    const outputNativeCurrency = this.swaps.find(({ outputAmount }) => outputAmount.currency.isNative)?.outputAmount
+      .currency
+
+    return {
+      inputAmount: this.inputAmount,
+      inputAmountNative: inputNativeCurrency
+        ? this.swaps.reduce((total, swap) => {
+            return swap.route.pathInput.isNative ? total.add(swap.inputAmount) : total
+          }, CurrencyAmount.fromRawAmount(inputNativeCurrency, 0))
+        : undefined,
```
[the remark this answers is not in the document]
zhongeric: - yes
- yes!

sdks/router-sdk/src/entities/trade.ts
```diff
@@ -155,13 +155,49 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
...
-      .map(({ outputAmount }) => outputAmount)
+      .map(({ outputAmount: routeOutputAmount }) => routeOutputAmount)
       .reduce((total, cur) => total.add(cur), CurrencyAmount.fromRawAmount(outputCurrency, 0))
 
     this._outputAmount = totalOutputFromRoutes
     return this._outputAmount
   }
 
+  /**
+   * Returns the sum of all swaps within the trade where the pathInput/Output is native
```
hensha256, line 166:
```suggestion
   * Returns the sum of all swaps within the trade
```

hensha256: maybe im wrong, but this function returns the sum of all swaps, even when pathInput/Output _isnt_ native?
Its just the 2 `xAmountNative` variables that might not get returned for this case

sdks/router-sdk/src/entities/trade.test.ts
```diff
@@ -627,6 +709,78 @@ describe('Trade', () => {
...
+      // Expect all route inputs to be ETH
+      expect(trade.swaps.every((swap) => swap.route.input.isNative)).toBe(true)
+      // However, expect the routes to be preserved (v2 using WETH and v4 using ETH)
+      expect(trade.swaps[0].route.pathInput).toEqual(weth)
+      expect(trade.swaps[1].route.pathInput).toEqual(ETHER)
+
+      // Expect inputAmountNative to correctly track only the amount required for the ETH input V4 route
+      expect(trade.amounts.inputAmountNative).toBeDefined()
+      expect(trade.amounts.inputAmountNative?.equalTo(amountv4)).toBe(true)
+      expect(trade.amounts.outputAmountNative).toBeUndefined()
```
hensha256, line 745: I notice that the new `amounts.inputAmount` and `amounts.outputAmount` are never tested
i guess i would additionally check in 1 test that they have the expected amounts too to be 100% sure

i.e. `trade.amounts.inputAmount` equals `v4Amount + v2Amount`? and I actually dont know what `outputAmount` would equal on an exactIn test is it the worst case slippage amount?

zhongeric: I added tests for inputAmount and outputAmount for these two test cases, and yeah inputAmount is the sum, outputAmount is hard to test for exact input like you said but its the same calculations as the original getter `trade.outputAmount`

zhongeric: Also added two more test cases

hensha256 (approved):

PR #301, opened 2025-02-07
fix(router-sdk): add support for inputs requiring wrap or unwraps

snreynolds:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups

### DIFF
--- a/sdks/router-sdk/src/entities/trade.ts
+++ b/sdks/router-sdk/src/entities/trade.ts
@@ -13,8 +13,8 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
   public readonly tradeType: TTradeType
   private _outputAmount: CurrencyAmount<TOutput> | undefined
   private _inputAmount: CurrencyAmount<TInput> | undefined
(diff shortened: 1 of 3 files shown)


sdks/router-sdk/src/entities/trade.ts
```diff
@@ -239,6 +273,11 @@ export class Trade<TInput extends Currency, TOutput extends Currency, TTradeType
     return new Percent(outputCurrency.wrapped.buyFeeBps.toNumber(), 10000)
   }
 
+  private isWrappedNative(currency: Currency): boolean {
```
hensha256, line 276: i guess this would ideally go in the Currency class? is there a reason it cant?

hensha256 (commented):
looks great! just one thought

sdks/router-sdk/src/entities/trade.ts
snreynolds: ideally yeah maybe it lives in the `Token` class but it would be a different PR bc its in sdk-core

PR #319, opened 2025-02-28
feat(smart-wallet): Initial commit

zhongeric:
## Description

Add base implementation for smart wallet SDK.

Notable changes:
- Added viem as a dep
- Bumped typescript to work with viem
- Added SmartWallet class, and encoder + callPlanner. 

SmartWallet:
- Main entrypoint for consumers
- Supports `encode*` and `create*` functions
- Encode returns `MethodParameters`, create returns `Call`

Helper classes:

CallPlanner:
- Take a series of Calls and abi encode them

## How Has This Been Tested?

Unit tests

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,77 @@
...
+  public static INTERFACE: Interface = new Interface(abi)
+
+  /**
+   * Creates method parameters for executing a batch of calls through a smart wallet
+   * @param calls Array of calls to encode
+   * @param options Execution options
+   * @param chainId The chain ID for the calls
+   * @returns Method parameters with calldata and value
+   */
+  public static encodeExecute(calls: Call[], options: ExecuteOptions = {}): MethodParameters {
```
snreynolds, line 26: execute options should be optional, or default to false? I think since the main use case is basically not an advanced mode, we should just expose a function thats the easiest to ingest from the wallet side

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,77 @@
...
+    const mode = this.getModeFromOptions(options)
+    const planner = new ExecuteCallPlanner()
+    for (const call of calls) {
+      planner.add(call.to, call.data, call.value)
+    }
+    // Add dummy opData for modes that require it
```
snreynolds, line 32: I'm confused on how this is used. Why do you need to loop through all the calls and add them to the planner. Can't you just initialize the planner with the array?

sdks/smart-wallet-sdk/src/smartWallet.ts
snreynolds: just giving some reasoning for why in the sdk doc it just accepted an array of calls. My thinking is just that starting with the simplest possible interface we could give to wallet seems like it makes the most since given 5792 can't even handle the advanced cases yet?

sdks/smart-wallet-sdk/src/types.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export interface AdvancedCall extends Call {
+  /** Whether to revert the entire transaction if this call fails */
+  revertOnFailure?: boolean
+}
+
+/**
+ * Parameters for method execution
+ */
+export interface MethodParameters {
```
snreynolds, line 26: I'm sure we have this defined in another sdk, maybe core?

sdks/smart-wallet-sdk/src/utils/callPlanner.ts
```diff
@@ -0,0 +1,52 @@
...
+  addAuthorize(_key: string): ExecuteCallPlanner {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * Add a command to revoke an operator
+   * @param operator The operator address to revoke
+   */
+  addRevoke(_operator: string): ExecuteCallPlanner {
+    throw new Error('Not implemented')
```
snreynolds, line 50: seems odd to put these in the planner? how come they arent defined at the top level?

sdks/smart-wallet-sdk/src/constants.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * ABI encoding for each mode type
+ */
+export const MODE_TYPE_ABI_ENCODING = {
+  [ModeType.BATCHED_CALL]: ['(address,uint256,bytes)[]'],
+  [ModeType.BATCHED_CAN_REVERT_CALL]: ['(address,uint256,bytes)[]'],
+  [ModeType.BATCHED_CALL_SUPPORTS_OPDATA]: ['(address,uint256,bytes)[]', 'bytes'],
+  [ModeType.BATCHED_CALL_SUPPORTS_OPDATA_AND_CAN_REVERT]: ['(address,uint256,bytes)[]', 'bytes']
```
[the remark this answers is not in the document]
zhongeric: We need some specification of how to encode the _executionData_ to be passed into _execute_ function though right?

sdks/smart-wallet-sdk/src/types.ts
zhongeric: It seems to be defined locally in v3, v4, universal-router-sdk :/ so maybe we can explore moving to sdk-core

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: It is optional in the current interface, my reasoning for including them even for the basic case is for the BATCHED_CALL_CAN_REVERT mode, which doesn't support partial reverts at the per call level (advancedCall which requires you to specify per call) but rather over the entire batch. Seems easy enough to add the flexibility for now

sdks/smart-wallet-sdk/src/constants.ts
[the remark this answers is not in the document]
snreynolds: yep sorry i realized that later forgot to update this comment, I still think we only need it defined twice?

BATCHED_CALL/BATCHED_CAN_REVERT_CALL - are the same
BATCHED_CALL_SUPPORTS_OPDATA/BATCHED_CALL_SUPPORTS_OPDATA_AND_CAN_REVERT - are the same

sdks/smart-wallet-sdk/src/types.ts
snreynolds: meh ok ya just leave it :P aha

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @returns Method parameters with calldata and value
+   */
+  public static encodeCalls(calls: Call[], options: ExecuteOptions = {}): MethodParameters {
+    const mode = this.getModeFromOptions(options)
+    if(mode != ModeType.BATCHED_CALL && mode != ModeType.BATCHED_CALL_CAN_REVERT) {
+      throw new Error(`Invalid mode: ${mode}`)
+    }
+    const planner = new CallPlanner(calls)
+    const data = ModeEncoder.encode(mode, planner)
+    const encoded = encodeFunctionData({
```
snreynolds, line 27: idea for later, but this could probably just be apart of ModeEncoder?

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,81 @@
...
+      data: methodParameters.calldata,
+      value: methodParameters.value
+    }
+  }
+
+  /**
+   * Get the mode type from the options
+   */
+  public static getModeFromOptions(options: ExecuteOptions): ModeType {
+    if(options.senderIsUser) {
```
snreynolds, line 68: revert with not supported if sender is not user

snreynolds: if the sender is the user, the mode should be either BATCHED_CALL_CAN_REVERT or BATCHED_CALL

sdks/smart-wallet-sdk/src/utils/callPlanner.ts
```diff
@@ -0,0 +1,73 @@
...
+      { type: 'bytes', name: 'data' },
+      { type: 'uint256', name: 'value' }
+    ]
+  }
+] as const
+
+/**
+ * CallPlanner is used to encode a series Calls
+ */
+export class CallPlanner {
```
snreynolds, line 20: nicee i like this

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: Ah oops good catch

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: yep, I removed the ModeEncoder and just put it into the top level

sdks/smart-wallet-sdk/src/utils/callPlanner.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    it('should sum the values of all calls', () => {
+      const planner = new CallPlanner([
+        { to: TEST_ADDRESS_1, data: TEST_DATA_1, value: TEST_VALUE_1 },
+        { to: TEST_ADDRESS_1, data: TEST_DATA_2, value: TEST_VALUE_2 }
+      ])
+      expect(planner.value.toString()).toBe('300')
+    })
+
+    it('should sum the values of all calls with bigint values', () => {
```
snreynolds, line 44: this test is the same as above no?

snreynolds (commented):
1 nit lgtm tho! just gotta get CI passing 🚢

snreynolds (approved):

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Main SDK class for interacting with ERC7821-compatible smart wallets
+ */
+export class SmartWallet {
+  /**
+   * Creates method parameters for executing a simple batch of calls through a smart wallet
+   * @dev does not support opData
+   * @param calls Array of calls to encode
+   * @returns Method parameters with calldata and value
+   */
+  public static encodeCalls(calls: Call[], options: ExecuteOptions = {}): MethodParameters {
```
Ayoakala, line 20: This looks like it applies to `/encode_5792`

Ayoakala: Will the new endpoint need to do anything special for `options` or can we keep this as the default?

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,92 @@
...
+    const executionData = planner.encode()
+    const encoded = this._encodeExecute(mode, executionData)
+    return {
+      calldata: encoded,
+      value: planner.value
+    }
+  }
+
+  /// To be implemented
+  public static encodeAdvancedCalls(calls: AdvancedCall[], opData: string, _options: ExecuteOptions = {}): MethodParameters {
```
Ayoakala, line 36: How would this sdk call percolate up to the Wallet? would these advanced calls be surfaced in the Client?

sdks/smart-wallet-sdk/src/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Call types for smart wallet calls
+ * Follows ERC-7579
+ */
+export enum ModeType {
+  BATCHED_CALL = '0x0100000000000000000000000000000000000000000000000000000000000000',
+  BATCHED_CALL_CAN_REVERT = '0x0101000000000000000000000000000000000000000000000000000000000000',
+  BATCHED_CALL_SUPPORTS_OPDATA = '0x0100000000007821000100000000000000000000000000000000000000000000',
+  BATCHED_CALL_SUPPORTS_OPDATA_AND_CAN_REVERT = '0x0101000000007821000100000000000000000000000000000000000000000000'
```
Ayoakala, line 16: For only advanced calls right? also would advanced calls be able to take in `BATCHED_CALL` and `BATCHED_CALL_CAN_REVERT`?

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -0,0 +1,92 @@
...
+  protected static _encodeBatchedCallSupportsOpdata(
+    planner: CallPlanner,
+    opData: `0x${string}`
+  ): `0x${string}` {
+    return encodeAbiParameters(
+      MODE_TYPE_ABI_PARAMETERS[ModeType.BATCHED_CALL_SUPPORTS_OPDATA],
+      [planner.encode(), opData]
+    )
+  }
+}
```
Ayoakala, line 92: Curious why `BATCHED_CALL_SUPPORTS_OPDATA` differs from the other modes? should we have this for `BATCHED_CALL_SUPPORTS_OPDATA_AND_CAN_REVERT` as well?

sdks/smart-wallet-sdk/src/types.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Represents a call to a smart contract
+ */
+export interface Call {
+  /** The address of the contract to call */
+  to: `0x${string}`
+  /** The encoded calldata for the call */
+  data: `0x${string}`
+  /** The amount of ETH to send with the call */
+  value: bigint
```
Ayoakala, line 10: are we using `bigint` to match viem types
image

sdks/smart-wallet-sdk/src/constants.ts
Ayoakala: Also should we include Batch of batches mode from 7821 (`0x01000000000078210002...`)?

Ayoakala (commented):
LGTM just some general questions

sdks/smart-wallet-sdk/src/types.ts
zhongeric: yep

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: BATCHED_CALL_SUPPORTS_OPDATA has different encoding since it supports bytes opData

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: you can pass in canRevert optionally into encodeCalls if you want

sdks/smart-wallet-sdk/src/constants.ts
zhongeric: yeah we'll include batch of batches in a later update when we flesh out AdvancedCalls

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: Not sure what you mean, this SDK will be used to generate the data which will be sent to the wallet over api?

snreynolds: I think for now we have nothing in the wallet that would use this feature tbh, its just a way though for the wallet to tell us to encode a call that can revert but not cause the transaction to fail

Ayoakala (approved):

PR #220, opened 2024-12-16
feat(permit2-sdk): separate cjs and esm builds, remove tsdx

just-toby:
## PR Scope

breaking change to build structure for `permit2-sdk`

## Description

updates the build setup for `permit2-sd` to manually and explicitly create separate cjs and esm builds (plus a types-only version)

## How Has This Been Tested?

built the sdk locally, then installed the local build result in another project (`universe/apps/web`) and verified that it properly resolved the esm version instead of the cjs version

## Are there any breaking changes?

there shouldn't be, but technically if an end user is importing incorrectly (by directly referencing any build artifact files) then they'll need to update

## (Optional) Follow Ups

 PRs to follow will do this for many other SDKs

PR #322, opened 2025-03-04
fix(permit2-sdk, universal-router-sdk): new build setup

just-toby:
## PR Scope

 change to build structure for `permit2-sdk` and `universal-router-sdk`

## Description

updates the build setup for these two SDKs to manually and explicitly create separate cjs and esm builds (plus a types-only version)

_**this is a copy of #220 because that was accidentally merged into a temporary graphite branch!!**_

## How Has This Been Tested?

built the sdk locally, then installed the local build result in another project (`universe/apps/web`) and verified that it properly resolved the esm version instead of the cjs version

there shouldn't be, but technically if an end user is importing incorrectly (by directly referencing any build artifact files) then they'll need to update


sdks/universal-router-sdk/src/entities/actions/unwrapWETH.ts
```diff
@@ -7,7 +7,7 @@ import { ROUTER_AS_RECIPIENT, WETH_ADDRESS } from '../../utils/constants'
 
 export class UnwrapWETH implements Command {
   readonly tradeType: RouterActionType = RouterActionType.UnwrapWETH
-  readonly permit2Data: Permit2Permit
+  readonly permit2Data?: Permit2Permit
```
adjkant, line 10: intentional change?

sdks/universal-router-sdk/src/index.ts
adjkant, line 1: Same on this file?

sdks/universal-router-sdk/src/entities/actions/unwrapWETH.ts
just-toby: yes, same as last PR. it fixes a typescript issue

sdks/universal-router-sdk/src/index.ts
just-toby: yes, same as last PR. it fixes a typescript issue

jsy1218 (approved):

PR #333, opened 2025-03-06
fix(sdk-core): revert to old build setup

just-toby:
## PR Scope

fix(sdk-core): revert build setup to use `tsdx` again :(

## Description

somewhere along the line when updating our build setup for all SDKs, we caused incompatibilities in downstream services (Routing API and Smart Order Router) because all services are using `instanceof` to check pool versions.

e.g. `pool instanceof V4Pool` 

this can fail if `V4Pool` is being imported in different contexts (cjs vs esm) because it will treat the two different constructors as different classes.

for now, we want to revert all the build changes to unblock unrelated feature work. in the future we can try to upgrade the build scripts here again, but only after the backend moves to using safer version checks like these typeguards:

```
function isV4Pool(pool: TPool): pool is V4Pool {
  return 'hooks' in pool && 'tickSpacing' in pool;
}

function isV3Pool(pool: TPool): pool is V3Pool {
  return 'fee' in pool && 'tickSpacing' in pool && !('hooks' in pool);
}

function isPair(pool: TPool): pool is Pair {
  return 'liquidityToken' in pool;
}
```

## How Has This Been Tested?

CI


jsy1218 (approved):

xrsv:
is this the PR you are reverting? https://github.com/Uniswap/sdks/pull/226/files
i see similar changes for v2/v3/v4-sdks, do we need to revert those as well?

jsy1218:
> is this the PR you are reverting? https://github.com/Uniswap/sdks/pull/226/files i see similar changes for v2/v3/v4-sdks, do we need to revert those as well?

@xrsv we do, one-by-one, unfortunately. We want to make sure v2-sdk revert includes the bumped sdk-core version, etc.

xrsv (approved):

PR #347, opened 2025-03-25
feat(UniswapX): Add expected amounts to DutchV3

codyborn:
## Description

Similar to Priority Trades, allow caller to pass in the expected amounts to the trade to override the amount in/out.

## How Has This Been Tested?

Unit tests

No.


Ayoakala (approved):
LGTM

PR #354, opened 2025-04-14
feat(breaking): Smart wallet upgrade

zhongeric:
# Changelog

## Breaking Changes
- **Removed `encodeCalls` method**: This method has been completely removed
- **New primary entrypoint**: `encodeBatchedCall` is now the main method for encoding call batches
  - Same parameter structure as the previous `encodeCalls` (accepts `calls[]` and `options`)
  - Returns the same `MethodParameters` object type as before
- **Default revert behavior changed**: The default value for `shouldRevert` in BatchedCallPlanner is now `true`, meaning batched calls will revert by default if any call fails

## Added
- Added `BatchedCallPlanner` class to handle batched calls with revert control
- Added `encodeERC7821BatchedCall` method for legacy support of the ERC-7821 specification
  - Use this only if you specifically need the ERC-7821 entrypoint
  - For all other cases, use the new `encodeBatchedCall` method

## Migration Guide
```typescript
// Before:
const result = SmartWallet.encodeCalls(calls, options);

// After:
const result = SmartWallet.encodeBatchedCall(calls, options);

// For ERC-7821 specific compatibility:
const result = SmartWallet.encodeERC7821BatchedCall(calls, options);
```

This update simplifies the API while maintaining backward compatibility through parameter structure. The main behavioral difference is that batched calls will now revert by default when any contained call fails, improving safety by preventing partial execution of transaction batches.

## How Has This Been Tested?


sdks/smart-wallet-sdk/README.md
```diff
@@ -16,10 +16,18 @@ or
 yarn add @uniswap/smart-wallet-sdk
 ```
 
+## Deployment Addresses
+
+Below is a table of contract deployment addresses across various networks:
+
+| Network | Address | Commit Hash |
+|---------|---------|------------|
+| Sepolia | 0x85479d2AebeFf8445769a265763666c0d3dDC508 | 732247c5e3146b9340cb29e0f2b8f9e2f1df67a4 |
```
snreynolds, line 25: should we tag commits with versions?

sdks/smart-wallet-sdk/abis/MinimalDelegationEntry.json
```diff
@@ -0,0 +1 @@
...o1oGLgR"],"license":"UNLICENSED"},"src/libraries/SignedBatchedCallLib.sol":{"keccak256":"0x5b46f93d6715cc29bdd5591eadbf7240390621cc89f183d16b550c2754487d3c","urls":["bzz-raw://f5cb725f79e019d6610ae3b73d50b8af423b631832be9f0592ae9df4e7a91230","dweb:/ipfs/QmQE7wM39w9CxDHk3xKoJ7YRwNEZUaQPHM2Fz2EZss3Dmp"],"license":"MIT"},"src/libraries/Static.sol":{"keccak256":"0xaf78f335887768df4267e8abf6f4acbd423c07f69d576bed84070cbcca35480c","urls":["bzz-raw://2d6c44dd8671b62275225ad520d8627af8928f03733e5e76e9f0e53b6cf3c76e","dweb:/ipfs/QmWGKvtYx9jC7RfdZCG6Avy9ZKPfALrYpSMFNGSY13uHPX"],"license":"UNLICENSED"},"src/libraries/TransientAllowance.sol":{"keccak256":"0x2dba244824723e9538c0c280726aad53a79aca876b589a04c368b31acab653cc","urls":["bzz-raw://6039868a862134a6e2a641be75b46a7759d5b867e6608c91545c6c770f60102e","dweb:/ipfs/QmRosmREFFhCNEV5WaNmFYgLd9Qq1qhSsdtwgy9uFpCMPs"],"license":"UNLICENSED"},"src/libraries/WrappedDataHash.sol":{"keccak256":"0x81d94e804efa4778c1f7098f9e7d9fda46b0fbc5fb149efcb323768fe3e4d4a4","urls":["bzz-raw://d950ad8309c651efe4f291db68343346b92a45d3198791ab6610bb3937226440","dweb:/ipfs/Qmch5Jk29XUivf5tJs2RboPMCU3pZqbrY2Vb5F6YEbJ4Cb"],"license":"UNLICENSED"}},"version":1},"id":41}
```
snreynolds, line 1: run 

`forge inspect MinimalDelegationEntry abi --json > MinimalDelegationEntry.json` and you'll just get the abi

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -60,16 +80,16 @@ export class SmartWallet {
...
     }
 
-    return ModeType.BATCHED_CALL
+    return ModeType.BATCHED_CALL_CAN_REVERT
   }
 
   /** Internal methods */
   
-  protected static _encodeExecute(mode: ModeType, data: `0x${string}`): `0x${string}` {
+  protected static _encodeERC7821Execute(mode: ModeType, data: `0x${string}`): `0x${string}` {
```
snreynolds, line 92: imo we should be encoding calldata as a direct call to execute not through the 7821 wrapper!

sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -1,40 +1,60 @@
...
    * @param calls Array of calls to encode
+   * @param options Basic options for the execution
    * @returns Method parameters with calldata and value
    */
-  public static encodeCalls(calls: Call[], options: ExecuteOptions = {}): MethodParameters {
+  public static encodeBatchedCall(calls: Call[], options: ExecuteOptions = {}): MethodParameters {
     const mode = this.getModeFromOptions(options)
     if(mode != ModeType.BATCHED_CALL && mode != ModeType.BATCHED_CALL_CAN_REVERT) {
       throw new Error(`Invalid mode: ${mode}`)
     }
```
snreynolds, line 25: So yeah I think this entrypoint should just be the direct encoding, not 7821 encoding

snreynolds: so it seems like you can just remove these lines/not support a mode being passed in

sdks/smart-wallet-sdk/src/smartWallet.ts
zhongeric: this is only supported for the legacy mode if desired, I marked it as deprecated do you think we should just remove it all together?

Ayoakala (approved):

sdks/smart-wallet-sdk/src/smartWallet.ts
snreynolds: I mean I can't see our internal code using this codepath but I think its fair to support in the sdk !

snreynolds (approved):

PR #378, opened 2025-07-24
fix(universal-router-sdk): fix mixed routes

dianakocsis:
## PR Scope

## Description

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -83,27 +83,79 @@ export class UniswapTrade implements Command {
...
+        return true
+      }
+
+      // if input currency is native and currency0 is not native, return true
+      // else return false.
+
+      // Ex:
+      // input is eth, first pool is [WETH, DAI]? - would need to wrap
+      // input is eth, first pool is [ETH, WETH]? - would not need to wrap
+      if (!(this.trade.swaps[0].route as unknown as V4Route<Currency, Currency>).pathInput.isNative) {
```
jsy1218: where is the currency0 check?

jsy1218:
```
if (this.trade.swaps[0].route as unknown as V4Route<Currency, Currency>).pathInput.isNative && this.trade.swaps[0].route.pools[0].currency0.isWrappedNative) {
    return true
}
```

ofc, `isWrappedNative` is not implemented, so we will need to implement it somewhere

dianakocsis: so this.trade.swaps[0].route covers a whole route
ex something like:
```
const v2Trade = new V2Trade(
        new RouteV2([WETH_USDC_V2, USDC_DAI], ETHER, DAI),
        CurrencyAmount.fromRawAmount(ETHER, inputEther),
        TradeType.EXACT_INPUT
)
const v3Trade = await V3Trade.fromRoute(
        new V3Route([WETH_DAI_V3, DAI_USDC_V3], ETHER, USDC),
        CurrencyAmount.fromRawAmount(ETHER, inputEther),
        TradeType.EXACT_INPUT
)
```
(this.trade.swaps[0].route as unknown as V4Route<Currency, Currency>).pathInput is WETH, pathOutput is DAI
(this.trade.swaps[1].route as unknown as V4Route<Currency, Currency>).pathInput is WETH, pathOutput is USDC

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -83,27 +83,79 @@ export class UniswapTrade implements Command {
...
+      // input is weth, first pool is [ETH, WETH]? - would not need to unwrap
+      if ((this.trade.swaps[0].route as unknown as V4Route<Currency, Currency>).pathInput.isNative) {
+        return true
+      }
+
+      return false
+    }
   }
 
   get outputRequiresWrap(): boolean {
```
jsy1218, line 123: dont forget to fix output wrap

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -81,44 +81,116 @@ export class UniswapTrade implements Command {
...
-    if (this.isAllV4) {
+    const swap = this.trade.swaps[0]
+    const lastRoute = swap.route
+    const lastPool = lastRoute.pools[lastRoute.pools.length - 1]
+
+    // if all pools are v4 or last pool is v4:
+    if (this.isAllV4 || lastPool instanceof V4Pool) {
```
snreynolds: think if (lastPool instanceof V4Pool) is equivalent here

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -81,44 +81,116 @@ export class UniswapTrade implements Command {
     return result
   }
 
+  // this.trade.swaps is an array of swaps / trades.
```
snreynolds, line 84: spoke offline about updating inputRequiresWrap to just check the following cases:

- first pool is a v4 pool: if the input is native & the first pool contains not native -> wrap
- else (first pool then must be a v2 or v3 pool) so just return inputCurrency.isNative

snreynolds: & inputRequiresUnwrap I believe can stay unchanged

snreynolds: sorry no we do need to update inputRequiredUnwrap but it just needs to change 1 line: We should just need to check if the first pool is a v4 pool (instead of the whole route is v4 pool) bc we could have a mixed route with the first pool being a v4 pool

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -81,44 +81,80 @@ export class UniswapTrade implements Command {
...
+
+    // if last pool is v4:
+    if (lastPool instanceof V4Pool) {
+      // If output currency is not native but path currency output is native, we need to wrap
       return (
         !this.trade.outputAmount.currency.isNative &&
-        (this.trade.swaps[0].route as unknown as V4Route<Currency, Currency>).pathOutput.isNative
+        (lastRoute as unknown as V4Route<Currency, Currency>).pathOutput.isNative
       )
+    } else {
```
snreynolds: simplify to just return false, dont need the else here bc we return early above

snreynolds (approved):

jsy1218:
i believe the output wrap change also works. before your output wrap fix, I see sim failure on my local routing-api https://app.warp.dev/block/zWcoB4tsbqkciVmkMRGsGC. afterwards, I see sim success https://app.warp.dev/block/Aep7Yd1rj675a3aYQtyFAo

PR #386, opened 2025-08-27
fix(smart-wallet-sdk): harden `createExecute` against prototype pollution attacks

ccashwell:
## Description

Hardened `SmartWallet.createExecute` against prototype pollution and invalid inputs; hardened the address map; added tests. This addresses a reported issue where passing inherited object keys like `"__proto__"` or `"constructor"` as `chainId` would incorrectly pass the existing check, producing malformed `to` addresses. Reported via [Cantina Finding #471](https://cantina.xyz/code/f9df94db-c7b1-434b-bb06-d1360abdd1be/findings?finding=471)

### Fix Details
- **Input validation** (`src/smart-wallet-sdk/src/smartWallet.ts`)
  - Normalize `chainId` (convert string to number when possible).
  - Validate it is a finite integer and a direct key of `SMART_WALLET_ADDRESSES` using `Object.prototype.hasOwnProperty.call`.
  - If invalid or absent, throw: `Smart wallet not found for chainId: ${chainId}`.
- **Map hardening** (`src/smart-wallet-sdk/src/constants.ts`)
  - Build `SMART_WALLET_ADDRESSES` with a null prototype via `Object.setPrototypeOf(map, null)` to block inherited property access (e.g., `__proto__`, `constructor`). Mark deprecated in favor of `getSmartWalletAddress`.
  - Export new `getSmartWalletAddress(chainIdLike)` which validates the provided chain id and either returns a valid address or throws.
- **Tests** (`src/smart-wallet-sdk/src/smartWallet.test.ts`)
  - Valid path: ensures `createExecute` returns expected `to`, `data`, `value` for a real numeric `ChainId`.
  - Invalid inputs: asserts throws for `"__proto__"`, `"constructor"`, arbitrary non-numeric strings (e.g., `"forty-two"`), and unsupported numeric IDs (e.g., `1337`).

### Security impact
- **Prevents prototype pollution vector** where inherited keys could be treated as valid `chainId` values, leading to malformed `to` addresses.

## How Has This Been Tested?

- New tests validate both happy path and error scenarios for `createExecute`.
- All package tests, including new additions, are green.

- No API changes. Behavior now throws for invalid or unsupported `chainId` values that previously might have slipped through due to inherited property lookups.

### DIFF
--- a/sdks/smart-wallet-sdk/src/smartWallet.ts
+++ b/sdks/smart-wallet-sdk/src/smartWallet.ts
@@ -3,7 +3,7 @@ import { encodeFunctionData } from 'viem'
 
 import abi from '../abis/MinimalDelegationEntry.json'
 
-import { ModeType, SMART_WALLET_ADDRESSES } from './constants'
+import { ModeType, getSmartWalletAddress } from './constants'
 import { Call, MethodParameters, ExecuteOptions } from './types'
 import { CallPlanner } from './utils'
 import { BatchedCallPlanner } from './utils/batchedCallPlanner'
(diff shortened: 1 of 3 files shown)


sdks/smart-wallet-sdk/src/smartWallet.ts
```diff
@@ -60,39 +60,42 @@ export class SmartWallet {
...
    * @returns The call to execute
    */
-  public static createExecute(methodParameters: MethodParameters, chainId: ChainId ): Call {
-    const address = SMART_WALLET_ADDRESSES[chainId]
-    if(!address) {
+  public static createExecute(methodParameters: MethodParameters, chainId: ChainId): Call {
+    // Normalize and validate chainId prior to lookup to avoid prototype pollution and inherited keys
+    const normalizedChainId = typeof (chainId as unknown) === 'string' ? Number(chainId) : chainId
+    const isValidNumericChainId =
+      typeof normalizedChainId === 'number' && Number.isFinite(normalizedChainId) && Number.isInteger(normalizedChainId)
```
zhongeric: Probably makes more sense for us to not expose the Object directly but a getter function which runs these checks (user provided key is a number, and the prototype pollution ones)?

ccashwell: done. note that i chose not to stop exporting `SMART_CONTRACT_ADDRESSES` in order to avoid interface changes, however i did mark it as deprecated for clarity.

zhongeric (approved):

scorzeth (approved):

PR #392, opened 2025-09-24
feat(flashtestations-sdk): add basic flashtestations-sdk directory structure, created using tsdx

Melvillian:
## PR Scope

## Description

This sdk will be used by unichain, flashbots, and other L2's to work with the onchain and offchain components of [Flashtestations](https://github.com/flashbots/flashtestations). Right now it does nothing, but in the future as more functionality is added it will reach a point where it can be 2 used for 2 purposes:

1. as an `import`able library that exposes a `verify_flashtestation_in_block` function to use for verifying Unichain blocks for the presence of a valid flashtestation
2. as a CLI with a `npx @uniswap/flashtestation-sdx verify <block_parameter> <workload_id>` to do the same thing as the importable library, but exposed through a CLI interfaces

## How Has This Been Tested?

I ran `yarn test` and verified all the tests passed

No

## (Optional) Feedback Focus

This PR doesn't expose any important functionality, but I want to break the sdk's work into multiple PR's so this is the first of them. So don't review this in detail, just look at the general directory structure to see if there are any high level problems with how I'm fitting this sdk into the greater `sdks/` repo

## (Optional) Follow Ups


sdks/flashtestations-sdk/src/index.ts
```diff
@@ -0,0 +1,6 @@
+export const sum = (a: number, b: number) => {
+  if ('development' === process.env.NODE_ENV) {
+    console.log('boop');
```
codyborn, line 3: Bold choice

sdks/flashtestations-sdk/src/crypto/hash.ts
```diff
@@ -0,0 +1,33 @@
+import { keccak256 as viemKeccak256, toBytes } from 'viem';
+
+/**
+ * Cross-platform keccak256 hash function using viem
+ * Supports both Node.js and browser environments
+ */
+export function keccak256(data: string | Uint8Array): `0x${string}` {
```
codyborn, line 7: nit: '0x' could be a constant

codyborn (approved):

PR #572, opened 2026-04-03
feat(ur-sdk): add SwapRouter.encodeSwaps() for encoding command-based routes

scorzeth:
## PR Scope

## Description

Introduces `SwapRouter.encodeSwaps(spec, swapSteps)`, a new entry point where routers provide explicit `SwapStep[]` plans and the SDK owns the safety envelope around them (permit/ingress, fees, final settlement, and exact-output refunds). This decouples route topology from SDK trade construction, adds input/invariant checks for the new API shape, supports UR 2.0/2.1.1 and `ApproveProxy`, and adds parity/integration coverage against `swapCallParameters` where the two flows are comparable.

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups


sqrlfirst (approved):

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -0,0 +1,165 @@
...
+      invariant(
+        !(step.type === 'V4_SWAP' && step.v4Actions.some(hasV4MaxHopSlippage)),
+        'MAX_HOP_SLIPPAGE_UNSUPPORTED_ON_V2_0'
+      )
+    }
+
+    switch (step.type) {
+      case 'V2_SWAP_EXACT_IN':
+      case 'V2_SWAP_EXACT_OUT':
+        assertRouterRecipient(step.recipient)
```
mgretzke, line 137: Always enforcing the router to be the recipient seems like a safer and better approach. However, I just want to flag that previously, routers could optimize for simple swaps (no WETH unwrapping in the end) and set the recipient directly as the user, which safes like 2.2k gas, because funds to not have to go to the router as an intermediary before sweeping.
Its a hyper optimization though and I don't know if routers actually did that, so feels worth it to me.

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -0,0 +1,165 @@
...
+  } else {
+    invariant(amountCurrency.equals(outputCurrency), 'INVALID_ROUTING_AMOUNT_CURRENCY')
+    invariant(quoteCurrency.equals(inputCurrency), 'INVALID_ROUTING_QUOTE_CURRENCY')
+  }
+
+  if (tokenTransferMode === TokenTransferMode.ApproveProxy) {
+    invariant(!!spec.chainId, 'PROXY_MISSING_CHAIN_ID')
+    invariant(!spec.routing.inputToken.isNative, 'PROXY_NATIVE_INPUT')
+    invariant(!spec.permit, 'PROXY_PERMIT_CONFLICT')
+    invariant(!!spec.recipient && recipient !== SENDER_AS_RECIPIENT, 'PROXY_EXPLICIT_RECIPIENT_REQUIRED')
```
mgretzke, line 97: Should we also validate `recipient !== ROUTER_AS_RECIPIENT` outside of the ApproveProxy block? If someone passes `ROUTER_AS_RECIPIENT` (address(2)) as spec.recipient in Permit2 mode, the map() function in Dispatcher.sol resolves it to address(this) (Universal Router).

I don't see a scenario where this could be the desired outcome, because it would mean the tokens would remain inside the router after the final sweep.

mgretzke: Additionally, we should talk about a `spec.recipient != address(0)` check. I believe this was unchecked in the old path as well, but except to burn to address(0), this is very likely a false input. If you explicitly wanted to burn the output, you could still use `0xdead` like we are doing with the firepit.

`spec.fee.recipient` falls in the same category, however way more trivial, so likely fine to skip a check

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -117,6 +137,107 @@ export abstract class SwapRouter {
...
+        getCurrencyAddress(inputToken),
+        maxAmountIn,
+        normalizedSpec.chainId!,
+        normalizedSpec.urVersion,
+        normalizedSpec.deadline
+      )
+    }
+
+    return SwapRouter.encodePlan(planner, inputToken.isNative ? maxAmountIn : BigNumber.from(0), {
+      deadline: normalizedSpec.deadline ? BigNumber.from(normalizedSpec.deadline) : undefined,
```
mgretzke, line 237: This is the same behavior as in the old path, but this essentially allows a zero input to pass as "no deadline". 
I am not sure if this breaks existing implementations currently building on this, but the safer approach would be to explicitly require a uint256.max input as "no deadline", instead of just accepting a zero input as such.

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -117,6 +137,107 @@ export abstract class SwapRouter {
...
+    )
+
+    // Assumes routers already normalized unused input into `routing.inputToken`.
+    if (normalizedSpec.tradeType === TradeType.EXACT_OUTPUT) {
+      planner.addCommand(
+        CommandType.SWEEP,
+        [getCurrencyAddress(inputToken), normalizedSpec.recipient, 0],
+        false,
+        normalizedSpec.urVersion
+      )
```
mgretzke, line 222: Nit: The old path had an opt-in safeMode that did a final SWEEP(ETH, recipient, 0) as a safety guard for any unexpected native ETH left in the router. Not critical since routers own their swap steps, but just noting that this is gone now and should be documented somewhere.

sdks/universal-router-sdk/src/utils/encodeSwapStep.ts
```diff
@@ -0,0 +1,55 @@
...
+      return
+    }
+    case 'V2_SWAP_EXACT_OUT': {
+      const params: any[] = [step.recipient, step.amountOut, step.amountInMax, step.path, false]
+      if (useV2_1_1) params.push(step.maxHopSlippage ?? [])
+      planner.addCommand(CommandType.V2_SWAP_EXACT_OUT, params, false, urVersion)
+      return
+    }
+    case 'V3_SWAP_EXACT_IN': {
+      const params: any[] = [step.recipient, step.amountIn, step.amountOutMin, step.path, false]
```
mgretzke, line 24: Since `payerIsUser` is hardcoded to false here, we have the same trade-off as with the other comment about enforcing ROUTER_AS_RECIPIENT: 
`payerIsUser = false` means the SDK always needs a separate PERMIT2_TRANSFER_FROM to pull tokens into the router first, whereas the old path could set payerIsUser = true and let the swap callback pull funds directly from the user via Permit2, saving the extra command and token transfer. Again small gas costs for a cleaner safety model (swap steps can never drain beyond maxAmountIn), but worth mentioning.

Combined with the other one mentioned, this path adds around 5k gas. With our post Fusaka gas prices on mainnet, this does not matter anymore (0.002 Cent currently). If gas prices were to go back to bull run peaks, this could add easily spike to over > 20 cents extra costs per swap. 
However, I have no idea if routers optimized for this before in the first place.

Again, I do not think this should change, just want to flag this.

sdks/universal-router-sdk/src/types/encodeSwaps.ts
```diff
@@ -0,0 +1,200 @@
...
+  amount: BigNumberish
+}
+
+export type UnwrapWeth = {
+  type: 'UNWRAP_WETH'
+  recipient: string
+  amountMin: BigNumberish
+}
+
+export type SwapStep = V2SwapExactIn | V2SwapExactOut | V3SwapExactIn | V3SwapExactOut | V4Swap | WrapEth | UnwrapWeth
```
mgretzke, line 85: Nit: Since SWEEP is not an allowed SwapStep type, the only sweeps are the SDK's final output SWEEP and the EXACT_OUTPUT input refund SWEEP. For EXACT_INPUT, input tokens should always be fully consumed by the swap steps. If, for some reason, tokens were to end up in the router outside of the expected flow (e.g., through a V4 hook), those tokens could not be swept back to the user. Same limitation existed in the old path.
Maybe worth documenting somewhere.

mgretzke: Actually, related to this but potentially more important: 
The SwapStep[] design appears to now enables route shapes the old RouterTrade couldn't express. For example, mixing exact-in and exact-out steps within a single route. Intermediate token leftovers from such routes also can't be swept for the same reason.
An example for such a route would be:
V3 exact-in producing as much USDC as possible, followed by a V4 exact-out consuming only part of this the USDC Intermediary token

mgretzke (commented):
Feels like a great improvement, very bullish on the general direction with this and well organized code.
Left a few comments on things I believe worth checking out, documenting, or at least being aware of.

sdks/universal-router-sdk/src/swapRouter.ts
mgretzke: Small correction: the behavior depends on how zero is passed in. deadline: 0 (number) is false and silently becomes "no deadline". But deadline: BigNumber.from(0) as an object should be true... this would actually pass 0 to the UR contract, where block.timestamp > 0 is always true, causing an instant revert.

sdks/universal-router-sdk/src/types/encodeSwaps.ts
scorzeth: Currently the idea is its the routers job is to make sure all tokens are end up either in input or output for the sweep done by the SDK. I added a comment above the function to make this more explicit, hopefully not too overloaded.

sdks/universal-router-sdk/src/swapRouter.ts
scorzeth: After some discussion I've added in safeMode, good flag

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
scorzeth: Added both
```
  invariant(recipient !== ZERO_ADDRESS, 'RECIPIENT_CANNOT_BE_ZERO')
  invariant(recipient !== ROUTER_AS_RECIPIENT, 'RECIPIENT_CANNOT_BE_ROUTER')
```
to the validation

sdks/universal-router-sdk/src/utils/encodeSwapStep.ts
scorzeth: Also added some mention of this to the comment above the main `encodeSwaps`

sdks/universal-router-sdk/src/utils/computeEncodeSwapsAmounts.ts
```diff
@@ -0,0 +1,51 @@
...
+import { encodeFee1e18, encodeFeeBips } from './numbers'
+import { SwapSpecification } from '../types/encodeSwaps'
+
+export type EncodeSwapsAmounts = {
+  maxAmountIn: BigNumber
+  grossMinOrExactOut: BigNumber
+  netMinOrExactOut: BigNumber
+}
+
+export function computeEncodeSwapsAmounts(spec: SwapSpecification): EncodeSwapsAmounts {
```
hensha256: would be good to have more comments in this PR. This whole file has 0 comments, and theres very few comments in the whole PR

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (variableSegmentLength < 23 || variableSegmentLength % 23 !== 0) return undefined
+
+  return variableSegmentLength / 23
+}
+
+function hasV4MaxHopSlippage(action: V4Action): boolean {
+  switch (action.action) {
+    case 'SWAP_EXACT_IN':
+    case 'SWAP_EXACT_OUT':
+      return action.maxHopSlippage !== undefined
```
hensha256:
```suggestion
```

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
hensha256: Yeah our standard UR SDK encoding does optimise for this - but we agreed that this new encoding style is significantly more complicated to catch all the edgecases, so better to keep it simpler and safer in this way

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+}
+
+export function validateEncodeSwaps(spec: SwapSpecification, swapSteps: SwapStep[]): void {
+  invariant(swapSteps.length > 0, 'EMPTY_SWAP_STEPS')
+
+  const recipient = spec.recipient ?? SENDER_AS_RECIPIENT
+  const tokenTransferMode = spec.tokenTransferMode ?? TokenTransferMode.Permit2
+  const urVersion = spec.urVersion ?? UniversalRouterVersion.V2_0
```
hensha256: worth noting that in the flow of `encodeSwaps` these lines are redundant because `encodeSwaps` runs `normalizeEncodeSwapsSpec`. Maybe it would be best for this function to take in a normalizedSpec as the parameter and remove the redudancy?

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -117,6 +142,113 @@ export abstract class SwapRouter {
...
+   * Router custody with `payerIsUser = false` is deliberate for safety, even if it can cost an extra
+   * command or transfer. The SDK does not infer route topology or add transition commands on behalf of routers.
+   */
+  public static encodeSwaps(spec: SwapSpecification, swapSteps: SwapStep[]): MethodParameters {
+    const normalizedSpec = SwapRouter.normalizeEncodeSwapsSpec(spec)
+    const planner = new RoutePlanner()
+
+    validateEncodeSwaps(normalizedSpec, swapSteps)
+
+    const { maxAmountIn, netMinOrExactOut } = computeEncodeSwapsAmounts(normalizedSpec)
```
hensha256: I feel like theres a disconnect on naming here.

```suggestion
    const { maxOrMaxAmountIn, netMinOrExactAmoutOut } = computeEncodeSwapsAmounts(normalizedSpec)
```

hensha256: also worth a comment about why `net` is included: this is post-fee/

sdks/universal-router-sdk/src/utils/encodeSwapStep.ts
hensha256: Again yes our old UR SDK functions do optimise for this, but better to keep it simple here agreed

sdks/universal-router-sdk/src/types/encodeSwaps.ts
```diff
@@ -0,0 +1,201 @@
...
+  safeMode?: boolean
+}
+
+export type V2SwapExactIn = {
+  type: 'V2_SWAP_EXACT_IN'
+  recipient: string
+  amountIn: BigNumberish
+  amountOutMin: BigNumberish
+  path: string[]
+  maxHopSlippage?: BigNumberish[]
```
sqrlfirst: @scorzeth @hensha256 I noticed that there is inconsistency between `Commands` in UniversalRouter contract and `swapSteps` in SDK regarding slippage representation. In swapSteps `maxHopSlippage` is used while in  UniversalRouter `minHopPriceX36` is [used](https://github.com/Uniswap/universal-router/blob/main/contracts/base/Dispatcher.sol#L178-L194). What do you think about it?

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
scorzeth: fixed

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
scorzeth: good call. cleaned this up and now have validate take a NormalizedSwapSpecification

sdks/universal-router-sdk/src/swapRouter.ts
scorzeth: ya made this more clear good call out

sdks/universal-router-sdk/src/utils/computeEncodeSwapsAmounts.ts
scorzeth: added substantial comments throughout

sdks/universal-router-sdk/src/utils/encodeV4Action.ts
```diff
@@ -0,0 +1,101 @@
+import { Actions, URVersion } from '@uniswap/v4-sdk'
+import { UniversalRouterVersion, isAtLeastV2_1_1 } from './constants'
+import { V4Action } from '../types/encodeSwaps'
+
+const ACTION_NAME_TO_ENUM: { [key: string]: Actions } = Object.fromEntries(
+  Object.entries(Actions).filter(([key]) => isNaN(Number(key)))
+) as { [key: string]: Actions }
+
+const V4_UR_VERSION_MAP: Record<string, URVersion> = {
```
dannythedawger, line 9: another dep for new UR versions to update! we could create a playbook for upgrading and keep it in the repo for context

sdks/universal-router-sdk/src/utils/encodeV4Action.ts
```diff
@@ -0,0 +1,101 @@
...
+  [UniversalRouterVersion.V2_0]: URVersion.V2_0,
+  [UniversalRouterVersion.V2_1_1]: URVersion.V2_1_1,
+}
+
+export function toV4URVersion(version?: UniversalRouterVersion): URVersion {
+  if (version === undefined) return URVersion.V2_0
```
dannythedawger, line 15: should we allow for an undefined version here? we are making a default decision and that imo should be on the caller of the sdk fucntion, not the sdk

dannythedawger (approved):
lgtm

PR #576, opened 2026-04-27
feat(sdk-core): add CurrencyAmount comparator overrides with currency…

shubhamessier:
# CurrencyAmount Comparison Overrides

Adds `lessThan`, `equalTo`, and `greaterThan` overrides on `CurrencyAmount`. Comparing amounts of different currencies now throws a `'CURRENCY'` error, matching the existing behavior of `add` and `subtract`.

The operand type is narrowed to `CurrencyAmount<T> | 0`, where `0` is a sentinel for zero comparisons. This rejects bare `BigintIsh` values at the type level.

Internal callers in `v2-sdk` and `router-sdk` are updated to use `0` in place of `JSBI.BigInt(0)`.

## Validation

All existing tests pass (22 + 103 + 338 + 240 + 360). Type checking and linting are clean.

## Changesets

- `@uniswap/sdk-core`: minor
- `@uniswap/v2-sdk`: patch
- `@uniswap/router-sdk`: patch

### DIFF
--- a/.changeset/currency-amount-comparators.md
+++ b/.changeset/currency-amount-comparators.md
@@ -1,0 +1,7 @@
+---
+"@uniswap/sdk-core": minor
+"@uniswap/v2-sdk": patch
+"@uniswap/router-sdk": patch
+---
+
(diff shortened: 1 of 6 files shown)


shubhamessier:
Hey @mikeki, would appreciate your review here.

cc: @ciamac @ccashwell

PR #593, opened 2026-05-28
fix(ur-sdk): Fix and improve v4 ur mapping

scorzeth:
## PR Scope

## Description

`toV4URVersion` is an anti-corruption layer between the UR-SDK and v4-sdk version models. This is to make it so devs don't ned to update a mapping when new Universal Router versions are released, just the one enum in `constants.ts`.

## How Has This Been Tested?

## (Optional) Feedback Focus

## (Optional) Follow Ups

### DIFF
--- a/sdks/universal-router-sdk/src/entities/actions/uniswap.ts
+++ b/sdks/universal-router-sdk/src/entities/actions/uniswap.ts
@@ -8,7 +8,6 @@ import {
   V4Planner,
   encodeRouteToPath as encodeV4RouteToPath,
   Actions,
-  URVersion,
 } from '@uniswap/v4-sdk'
 import {
   Trade as RouterTrade,
@@ -26,6 +25,7 @@ import {
 } from '@uniswap/router-sdk'
 import { Permit2Permit } from '../../utils/inputTokens'
 import { getPathCurrency } from '../../utils/pathCurrency'
+import { toV4URVersion } from '../../utils/toV4URVersion'
 import { Currency, TradeType, Token, CurrencyAmount, Percent } from '@uniswap/sdk-core'
 import { Command, RouterActionType, TradeConfig } from '../Command'
 import {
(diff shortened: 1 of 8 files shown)


kristiehuang (approved):
Lol ty!! think i tried to improve this with toV4URVersion but maybe didnt catch all the places

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -495,7 +481,7 @@ function addV4Swap<TInput extends Currency, TOutput extends Currency>(
   const perHopSlippage = minHopPriceX36?.map((s) => BigNumber.from(s)) ?? []
 
   const v4Planner = new V4Planner()
-  v4Planner.addTrade(trade, slippageToleranceOnSwap, perHopSlippage, toURVersion(options.urVersion))
+  v4Planner.addTrade(trade, slippageToleranceOnSwap, perHopSlippage, toV4URVersion(options.urVersion))
```
dannythedawger, line 484: I assume this is well covered by tests?

sdks/universal-router-sdk/src/utils/toV4URVersion.ts
```diff
@@ -0,0 +1,14 @@
+import { URVersion } from '@uniswap/v4-sdk'
```
dannythedawger, line 1: this should really be held here instead of the v4 sdk. I know the ur sdk depends on the v4 sdk, so v4 can't import it. curious tho why v4 sdk needs the ur versions?

dannythedawger (commented):
great cleanup, left some qs

sdks/universal-router-sdk/src/utils/toV4URVersion.ts
scorzeth: v4-sdk needs it for knowing if it needs to encode minHopPriceX36 or not based on version

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
scorzeth: we have [toV4URVersion.test.ts](https://github.com/Uniswap/sdks/pull/593/changes#diff-95cf1b6784bf1fea818d950d2848c753721398d759cb7a8aebc054ff78d467d5) covering the direct version mapping tests and `test/unit/perHopSlippage.test.ts` has tests that cover this path

dannythedawger (approved):
lgtm

PR #597, opened 2026-06-04
feat(sdk-core): add Robinhood (4663) and Arc (5042) chain support

jsy1218:
## Description

Adds deployment-address support for two new chains — **Robinhood (4663)** and **Arc (5042)** — across `sdk-core` and `universal-router-sdk`, so routing and Universal Router usage work on them. Mirrors the MegaETH pattern from #585.

Address data sourced from `Uniswap/contracts` deployments (`4663.md`, `5042.md`) and UniRoute's hardcoded chain configs.

### sdk-core
- **`chains.ts`**: corrected `ChainId.ROBINHOOD` from `46630` → **`4663`** (matches the canonical Robinhood mainnet chain ID used in the Uniswap backend `config.ts`, `unirpc-v2`, the `CHAIN_ROBINHOOD` proto, and `contracts/deployments/4663.md`); added `ARC = 5042`; added both to `SUPPORTED_CHAINS`; registered Arc's average block time.
- **`addresses.ts`**: added Robinhood/Arc to `V2_FACTORY_ADDRESSES` and `V2_ROUTER_ADDRESSES`; added `ARC_ADDRESSES` and `ROBINHOOD_ADDRESSES` (`ChainAddresses` with the full V2/V3/V4 set); registered both in `CHAIN_TO_ADDRESSES_MAP`. All derived maps (quoter, multicall, tick lens, NPM, swap router 02, v3 factory) auto-populate.
- **`entities/weth9.ts`**: added wrapped-native entries for `4663` (`0x0bd7d308…`) and `5042` (`0x8bceaa40…`). Arc's native currency is USDC; the `5042` entry is the wrapped-native token used in Arc's on-chain contract deployments.

### universal-router-sdk
- **`constants.ts`**: added `CHAIN_CONFIGS[4663]` and `CHAIN_CONFIGS[5042]`, each with the V2_1_1 Universal Router and `swapProxy`.

## How Has This Been Tested?

- `bun run --filter @uniswap/sdk-core test` → **372 pass** (was 370).
- `bun run --filter @uniswap/sdk-core build` → typecheck/build clean (confirms both chains satisfy `Record<SupportedChainsType, ChainAddresses>`).
- `hardhat test test/utils/constants.test.ts` in `universal-router-sdk` → **162 passing** (was 154), covering UR address/creationBlock for 4663 and 5042 and the `V1_2`-not-deployed throws. (Hardhat then emits the same benign `relative URL without a base` post-test error noted in #585.)

[...]

PR #599, opened 2026-06-05
fix(chains): fix Arc and Robinhood chain constants

cartcrom:
## Description

Corrects the Arc and Robinhood constants that landed in #597 before the pending changeset is released.

## Changes

- Sets Arc average block time to `0.48` seconds.
- Treats Arc as having no IWETH9-compatible wrapped native token:
  - removes `WETH9[5042]`
  - uses `WETH_NOT_SUPPORTED_ON_CHAIN` in the Universal Router config
- Replaces placeholder Universal Router creation blocks with exact deployment blocks:
  - Arc: `1950059`
  - Robinhood: `18127`
- Keeps Robinhood WETH configured as `0x0Bd7D308f8E1639FAb988df18A8011f41EAcAD73`.
- Adds tests for Arc no-WETH behavior, exact UR addresses/creation blocks, and unsupported UR versions.

## Why

Arc’s deployment used `UnsupportedProtocol` for WETH-like constructor slots. That address should not be exposed as a real WETH9 token because wrap/unwrap paths would be constructed against a contract that intentionally does not support them. This matches the existing Celo/Tempo pattern of using `WETH_NOT_SUPPORTED_ON_CHAIN`.

## Validation

- `git diff --check` passed.
- `sdks/sdk-core/node_modules/typescript/bin/tsc -p sdks/sdk-core/tsconfig.cjs.json --noEmit` passed.

Could not run Bun package tests locally because this shell does not have the `bun` binary installed (`command not found: bun`).

### DIFF
--- a/sdks/sdk-core/src/chains.test.ts
+++ b/sdks/sdk-core/src/chains.test.ts
@@ -6,7 +6,7 @@ describe('getAverageBlockTimeSecs', () => {
     expect(getAverageBlockTimeSecs(ChainId.ARBITRUM_ONE)).toEqual(0.25)
     expect(getAverageBlockTimeSecs(ChainId.ROBINHOOD)).toEqual(0.1)
     expect(getAverageBlockTimeSecs(ChainId.MEGAETH)).toEqual(1)
-    expect(getAverageBlockTimeSecs(ChainId.ARC)).toEqual(1) // TODO(arc): confirm average block time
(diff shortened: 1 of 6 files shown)


matteenm (approved):

dannythedawger (approved):
lgtm

jsy1218 (approved):

PR #608, opened 2026-06-10
feat(universal-router-sdk): add nativeErc20Input option for native-ERC20 gas tokens

cartcrom:
## Motivation

Some chains expose their **native gas token through an ERC20 predeploy** whose balance mirrors the native balance 1:1. On Arc (chain 5042), USDC is the gas token: the native side is 18 decimals, and a 6-decimal ERC20 interface lives at the `0x3600…0000` predeploy. Arc has no WETH (already `WETH_NOT_SUPPORTED_ON_CHAIN` in this SDK).

Today the SDK can only encode swaps from such a token as a plain ERC20 input, forcing users through a Permit2 approval + permit signature **even though they hold the chain's gas token**. But the Universal Router can handle this without any approval:

1. send the tx with the input attached as `{value}` in native units
2. encode the route against the ERC20 input
3. `payerIsUser = false`, so the router pays pools from its own (msg.value-funded) balance
4. never wrap — the ERC20 interface is directly usable

## What this adds

A new opt-in `SwapOptions` flag, **`nativeErc20Input`**, on the `SwapRouter.swapCallParameters` path:

- tx `value = maximumAmountIn × 10^(18 − token.decimals)` (×10¹² for Arc USDC; ×1 for an 18-decimal case like CELO)
- `payerIsUser = false` on V2/V3 swap commands and the V4 `SETTLE`
- no `WRAP_ETH` or `PERMIT2_PERMIT` commands
- unused input swept back to the recipient (as the ERC20) on exact-output / partial-fill-risk trades, mirroring the existing ApproveProxy refund
- rejects misuse: native input, `inputTokenPermit`, `TokenTransferMode.ApproveProxy`, native-keyed V4 `pathInput`, and tokens with >18 decimals

**Off by default — when the flag is unset every existing code path is unchanged** (the `payerIsUser` branch, value computation, and refund block all fall through identically).

## Validation

- 11 new decode-level unit tests in `test/unit/nativeErc20Input.test.ts` (value scaling, exact command bytes, decoded `payerIsUser` on V3 + V4 `SETTLE`, exact-output sweep, all rejection paths, and a default-off regression case).
- **Validated live on Arc**: this exact encoding runs end-to-end through the Uniswap trading API against Arc — gas estimation and execution succeed, confirming the predeploy credits msg.value into the router's ERC20 balance.
- The forge interop suite is untouched (no new `interop.json` fixtures): a mainnet fork can't execute this path since attached ETH doesn't credit a mainnet ERC20 — the live Arc validation covers execution instead.

## Design note for reviewers

The flag is **caller-asserted** (the integrator declares "this token is the chain's native-ERC20 gas token"), matching the `useRouterBalance` precedent. The alternative is an SDK-owned chain→predeploy map, which is safer against misuse (setting the flag on a normal ERC20 would strand the attached value in the router) at the cost of a hardcoded registry. Happy to switch if a map is preferred.

Changeset included (minor). Context: Arc needs this today, and CELO could adopt the same path (its native token is likewise an ERC20).

### DIFF
--- a/sdks/universal-router-sdk/src/entities/actions/uniswap.ts
+++ b/sdks/universal-router-sdk/src/entities/actions/uniswap.ts
@@ -57,6 +57,15 @@ export enum TokenTransferMode {
 
 export type SwapOptions = Omit<RouterSwapOptions, 'inputTokenPermit'> & {
   useRouterBalance?: boolean
+  /**
+   * The input Token is the chain's native gas token exposed via an ERC20 predeploy whose balance
+   * mirrors the native balance (e.g. USDC on Arc). The swap is funded by attaching
+   * msg.value = maximumAmountIn * 10^(18 - token.decimals) instead of pulling via Permit2:
+   * the router self-funds (payerIsUser = false) and unused input is swept back to the recipient
+   * on exact-output / partial-fill-risk trades. No ERC20 approval or permit is ever needed.
+   * Incompatible with native input, inputTokenPermit, and TokenTransferMode.ApproveProxy.
+   */
+  nativeErc20Input?: boolean
   inputTokenPermit?: Permit2Permit
   flatFee?: FlatFeeOptions
   safeMode?: boolean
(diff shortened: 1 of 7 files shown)


cartcrom:
Added `encodeSwaps` support in d361a0c2 — `nativeErc20Input` is now on `SwapSpecification` as well.

Turned out most of the machinery is already inherent to that flow (steps run with router custody / `payerIsUser = false` hardcoded, and the exact-output input refund sweep is unconditional), so the change is just: the flag on `SwapSpecification`, the same misuse rejections in `validateEncodeSwaps` (native input / permit / ApproveProxy / >18-decimal tokens), skipping the `PERMIT2_TRANSFER_FROM` ingress, and attaching `msg.value = exactOrMaxAmountIn × 10^(18 − token.decimals)` at `encodePlan`. No wrap handling needed since routers own `WRAP_ETH` steps and won't emit one for this path.

Eight new unit tests in `encodeSwaps.test.ts` cover the ingress skip, value scaling (6- and 18-decimal), the exact-output refund, all rejection paths, and a default-off regression (`PERMIT2_TRANSFER_FROM` + zero value when unset).

scorzeth (approved):
lgtm

gretzke (approved):

sdks/universal-router-sdk/src/entities/actions/uniswap.ts
```diff
@@ -345,6 +374,14 @@ export class UniswapTrade implements Command {
           this.options.recipient,
           0,
         ])
+      } else if (this.options.nativeErc20Input) {
+        // msg.value funded the router with maximumAmountIn (native balance == ERC20 balance via predeploy);
+        // sweep any unused input ERC20 back to the user
+        planner.addCommand(CommandType.SWEEP, [
+          getCurrencyAddress(this.trade.inputAmount.currency),
```
hensha256, line 381: this line imo should sweep using `ETH_ADDRESS` (and sweep as native), instead of sweeping the ERC20 balance. On Arc, ERC20 has 6 decimals and Native has 18 decimals so you can end up leaving dust behind in the router if you do not reference the currency with the full 18 decimals.

hensha256: you can combine this `else if` with the one below as they should now have the same SWEEP params

cartcrom: Good catch — fixed in #610 (sweeps `ETH_ADDRESS` and merges the two branches as suggested; also applied the same fix to the `encodeSwaps` exact-output refund, which had the identical dust issue).

PR #610, opened 2026-06-10
fix(universal-router-sdk): sweep nativeErc20Input refunds as native, not ERC20

cartcrom:
Follow-up to #608, addressing @hensha256's post-merge review comments ([1](https://github.com/Uniswap/sdks/pull/608#discussion_r3391808140), [2](https://github.com/Uniswap/sdks/pull/608#discussion_r3391812526)):

> this line imo should sweep using `ETH_ADDRESS` (and sweep as native), instead of sweeping the ERC20 balance. On Arc, ERC20 has 6 decimals and Native has 18 decimals so you can end up leaving dust behind in the router if you do not reference the currency with the full 18 decimals.

> you can combine this `else if` with the one below as they should now have the same SWEEP params

## Changes

- **`UniswapTrade.encode()`** — the `nativeErc20Input` exact-output / partial-fill-risk refund branch is merged into the existing native-input branch: both now `SWEEP(ETH_ADDRESS, recipient, 0)`. The router's leftover lives in its 18-decimal native balance; the previous ERC20 sweep floored to the token's decimals (6 on Arc) and could strand sub-decimal dust.
- **`SwapRouter.encodeSwaps()`** — same fix on the parallel path: the exact-output refund sweeps `ETH_ADDRESS` instead of the input ERC20 when `nativeErc20Input` is set (the dust reasoning applies identically there; flagging since the original comments were on the `swapCallParameters` path only).
- Updated the two affected unit tests to assert the native sweep; patch changeset included.

Full unit suite passes (141 tests).

### DIFF
--- a/.changeset/native-erc20-input-sweep-native.md
+++ b/.changeset/native-erc20-input-sweep-native.md
@@ -1,0 +1,5 @@
+---
+'@uniswap/universal-router-sdk': patch
+---
+
+`nativeErc20Input` exact-output / partial-fill refunds now sweep the leftover input as native (`ETH_ADDRESS`) instead of the ERC20. The router's leftover lives in its native balance (18 decimals); an ERC20 sweep floors to the token's decimals (e.g. 6 for Arc USDC) and can strand sub-decimal dust in the router.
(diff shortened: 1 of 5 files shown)


hensha256 (approved):

scorzeth (approved):
lgtm

PR #638, opened 2026-07-13
feat(ur-sdk): allow direct transfers as an option on swapsteps

scorzeth:
## Description

`SwapRouter.encodeSwaps` gains `allowDirectTransfers` — an opt-in validation regime that lets routing-supplied steps move funds directly between the user and pools instead of forcing every transfer through router custody. For a simple swap this removes both ERC-20 round-trips through the router (the Permit2 ingress and the final sweep transfer).

The safety contract is identical in both regimes: **on any successful transaction the user pays at most `exactOrMaxAmountIn` and the recipient receives at least `netMinOrExactAmountOut`** — any plan that could violate either is rejected at encode time or reverts on-chain. (Direct-output coverage additionally tolerates a bounded, sub-economic rounding shortfall — see Outbound coverage.) With the flag on, this is enforced by two equations instead of custody structure:

- **Inbound budget** — every direct pull (`payerIsUser` on V2/V3 swaps and v4 `SETTLE`; v4 `SETTLE_ALL`) must be a concrete uint160 amount of the input token, and their contract-enforced maxima must sum within `exactOrMaxAmountIn`. The SDK ingress pulls only the remainder (omitted at zero).
- **Outbound coverage** — step recipients are limited to the router or the spec recipient (`TAKE_ALL` only behind the sender sentinel; `WRAP_ETH` stays router-only), and contract-enforced direct-output minimums reduce the final `SWEEP` floor, which always remains as the residual forwarder. The floor tolerates a bounded sub-economic rounding shortfall — `max(min(0.5bps of the trade minimum, 1% of slippage), a small per-leg allowance)`, zero at zero slippage — so a fully-covered direct plan isn't reverted by a few wei of independent per-leg flooring; a larger, real gap still enforces the full floor.

Portion fees force full output custody (`PAY_PORTION` must observe the whole output). Protective amounts always derive from the spec — step amounts are only counted or capped, never trusted.

Also exports `computeEncodeSwapsAmounts` so integrators can derive trade-level amount bounds without reimplementing the math.

## How Has This Been Tested?

80 unit tests covering every invariant's reject/accept boundary, decoded-calldata assertions for the encoder changes, and flag-on/off byte-identity for custody plans, plus direct unit tests for the sweep-floor rounding tolerance (`directTransferSweepFloor`). A real-routing corpus fixture (production swapSteps shapes: hex paths, `CONTRACT_BALANCE` sentinels, mixed v4 action lists) pins cross-regime byte-identity and the `hookData` rejection. Full package unit suite green (465 passing); legacy `encodeSwaps` tests untouched.

No working caller changes. Two deliberate notes:

- v4 `SETTLE_ALL` / `TAKE_ALL` are now rejected at encode time unless `allowDirectTransfers` is set — they act on `msg.sender` directly and could previously only revert, double-charge, or bypass the fee/sweep envelope (shipped as minor; see changeset).
- A zero-amount exact-input swap now omits the ingress command instead of encoding a zero-amount `PERMIT2_TRANSFER_FROM` (on-chain no-op either way; noted for calldata-diff tooling).


sdks/universal-router-sdk/src/utils/directTransfers.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// contract-guaranteed amounts a step delivers directly to `recipient`
+function stepDirectOutputCredits(step: SwapStep, recipient: string): DirectOutputCredit[] {
+  const isDirect = 'recipient' in step && step.recipient.toLowerCase() === recipient.toLowerCase()
+  switch (step.type) {
+    case 'V2_SWAP_EXACT_IN':
+      return isDirect ? [{ token: step.path[step.path.length - 1], minAmount: BigNumber.from(step.amountOutMin) }] : []
+    case 'V2_SWAP_EXACT_OUT':
+      return isDirect ? [{ token: step.path[step.path.length - 1], minAmount: BigNumber.from(step.amountOut) }] : []
```
gretzke: This credit isn't contract-enforced. `v2SwapExactOutput` only checks the precomputed nominal input against `amountInMaximum` and then `_v2Swap` recomputes the actual output from what the pair really received. There is no post-swap assertion that the recipient got `amountOut`, unlike `v2SwapExactInput` (recipient balance delta check) and unlike V3 exact-out (`V3InvalidAmountOut`).

So with a fee-on-transfer input or intermediate token, a direct V2 exact-out leg can deliver less than its nominal amount and still succeed. Concrete case: net floor 90, this leg is credited 90 and drops the sweep floor to 0, a FoT input token makes it deliver 82, the tx succeeds and the recipient gets 82. In custody mode the same route reverts because the sweep checks the router's actual balance. That makes this the one spot where the flag-on regime is strictly weaker than custody on a successful tx.

The PR already applies the right standard to v4 exact-out (not asserted by the router, so it contributes zero). I think V2 exact-out needs the same treatment: credit zero here, and flip the "counts V2 exact-out amounts as credits" test. The leg can still pay the recipient directly, the sweep floor just stays at full netMin so the unverifiable portion has to be covered by custody flow.

sdks/universal-router-sdk/src/swapRouter.ts
```diff
@@ -214,9 +225,18 @@ export abstract class SwapRouter {
     }
 
     // Assumes routers already normalized final gross output into `routing.outputToken`.
+    // Direct-output minimums (contract-enforced, delivered straight to the recipient)
+    // reduce the floor; the sweep always remains to forward any custody balance.
+    const directOutputMin = normalizedSpec.allowDirectTransfers
```
gretzke: Worth a sentence in the doc comment that these credits are nominal amounts sent, not amounts received. For a fee-on-transfer output token, custody mode fails safe at the sweep (the floor is checked on the router's post-fee balance) while direct mode can succeed short. Declaring FoT tokens out of scope is probably the right call, but it should be written down.

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -51,24 +46,53 @@ function assertRouterRecipient(recipient: string): void {
...
 }
 
-function assertRouterActionRecipient(recipient: string): void {
-  invariant(recipient === ROUTER_AS_RECIPIENT, 'V4_ACTION_RECIPIENT_MUST_BE_ROUTER')
+// `routerOnlyError` is the surface's legacy flag-off code (steps vs v4 actions)
+function checkRecipient(spec: NormalizedSwapSpecification, recipient: string, routerOnlyError: string): void {
+  invariant(typeof recipient === 'string', routerOnlyError)
+  if (recipient === ROUTER_AS_RECIPIENT) return
+  invariant(spec.allowDirectTransfers, routerOnlyError)
+  invariant(spec.fee?.kind !== 'portion', 'PORTION_FEE_REQUIRES_ROUTER_CUSTODY')
```
gretzke, line 54: Portion fees force custody but flat fees don't, and a flat fee combined with a route that sends all output directly leaves nothing in custody to pay the fee TRANSFER. On a clean router that reverts, which is safe, but it's a liveness footgun for integrators. Might deserve a mention in the encodeSwaps docs, or a validation that a flat-fee spec keeps at least one custody output leg.

.changeset/allow-direct-transfers.md
```diff
@@ -0,0 +1,9 @@
+---
+'@uniswap/universal-router-sdk': minor
+---
+
+`SwapRouter.encodeSwaps`: add `allowDirectTransfers`, an opt-in validation regime letting routing-supplied steps move funds directly between the user and pools (per-step `payerIsUser` on V2/V3 swaps and v4 SETTLE, step recipients equal to the spec recipient, v4 SETTLE_ALL/TAKE_ALL). The SDK enforces an inbound budget (total contract-enforced direct pulls never exceed `exactOrMaxAmountIn`; ingress pulls only the remainder) and outbound coverage (contract-enforced direct output minimums reduce the final sweep floor). Portion fees still require full output custody.
+
+Tightened in the default regime: v4 `SETTLE_ALL` and `TAKE_ALL` are now rejected at encode time unless `allowDirectTransfers` is set — both act on `msgSender()` directly on-chain and could previously only revert, double-charge, or bypass the fee/sweep envelope under custody encoding.
+
+Also exports `computeEncodeSwapsAmounts` (and the `EncodeSwapsAmounts` type) so integrators can derive the trade-level amount bounds from a `SwapSpecification` without reimplementing the math.
```
gretzke: Small wording thing: this says the bounds are derived from a `SwapSpecification`, but `computeEncodeSwapsAmounts` takes a `NormalizedSwapSpecification`, so integrators need `normalizeEncodeSwapsSpec` first. Worth saying here or in the export's doc comment.

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -51,24 +46,53 @@ function assertRouterRecipient(recipient: string): void {
...
       case 'TAKE':
       case 'TAKE_PORTION':
-        assertRouterActionRecipient(action.recipient)
+        // fees are sdk-authored (the envelope's PAY_PORTION); a step-level TAKE_PORTION never pays the fee recipient
+        checkRecipient(spec, action.recipient, 'V4_ACTION_RECIPIENT_MUST_BE_ROUTER')
+        break
+      case 'TAKE_ALL':
+        // TAKE_ALL pays msgSender on-chain; any other spec recipient would be the
+        // wrong payee while still reducing their sweep floor
+        // flag-off TAKE_ALL is already rejected by the safe-mode gate above, so only a portion fee can block it here
```
gretzke: This case relies on the caller's flag-off gate having already run, which the comment acknowledges. A redundant `invariant(spec.allowDirectTransfers, ...)` here would keep the function safe under future refactors at zero cost.

gretzke (changes requested):

sdks/universal-router-sdk/src/utils/directTransfers.ts
scorzeth: Fixed — V2 exact-out now contributes zero, matching the standard we already apply to v4 exact-out

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
scorzeth: Done — added invariant(spec.allowDirectTransfers, 'TAKE_ALL_REQUIRES_DIRECT_TRANSFERS') on the TAKE_ALL path so it stays safe on its own if the safe-mode gate is ever moved.

.changeset/allow-direct-transfers.md
scorzeth: Fixed — the changeset now notes computeEncodeSwapsAmounts takes a NormalizedSwapSpecification, so integrators run normalizeEncodeSwapsSpec on a SwapSpecification first.

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
scorzeth: Documented rather than validated, since it fails safe. Added a note on encodeSwaps that a flat fee is a plain TRANSFER, so routing must keep at least the fee amount in router custody (i.e. not route 100% of output directly) or the TRANSFER reverts. I chose docs over a static Σ custody output ≥ fee check because it fails closed (reverts, no fund loss) and the check adds surface for a routing-owned concern — but happy to add the validation instead if you'd rather it be enforced.

sdks/universal-router-sdk/src/swapRouter.ts
scorzeth: Reframed rather than declared out of scope. Output minimums bind the amount each swap produces (pre-transfer) — which V3/V4 assert on the pool/ledger delta — not the recipient's post-tax balance. So a fee-on-transfer output token's own transfer tax reducing the final balance below the minimum is inherent to the token and expected when swapping into one.

Two things keep it sound:
 - The genuinely-unasserted leg (V2 exact-out) is credited zero (other thread), so we never credit a produced amount the router doesn't assert.
 - A FoT-aware router can still get a true received guarantee with no SDK special-casing, by encoding the tax into the amounts: a V2 exact-in terminal leg (whose amountOutMin is a recipient balance-delta check = post-tax received), or a V3/V4 leg with the min inflated by 1/(1−fee) so produced × (1−fee) ≥ target. Both reconcile with the credit + sweep-floor logic as-is.

Updated the JSDoc caveat to say this.

ericneil-sanc (approved):

.changeset/allow-direct-transfers.md
```diff
@@ -0,0 +1,9 @@
+---
```
dannythedawger: do we want to check this in?

gretzke (approved):

dannythedawger (approved):
lgtm! a some tests we should add

- Flat fee + 100% direct output is rejected (or pinned): a flat fee with a single direct exact-out leg paying the full output to the recipient must not encode into calldata whose fee TRANSFER always reverts on-chain — test/unit/directTransfers.test.ts (next to the existing flat-fee split test at line ~1121).
- Flat fee funding boundary is accepted: direct legs covering netMin plus a custody leg whose minimum ≥ fee.amount still encodes successfully — test/unit/directTransfers.test.ts.
- Zero-minimum direct legs don’t widen the rounding tolerance: a plan with one real direct leg short by 4 wei plus two amountOutMin: 0 direct legs must still enforce the full sweep floor (they add legs but no coverage today) — test/unit/directTransfers.test.ts (in the directTransferSweepFloor describe, ~line 967).
- Exact-output refund with a user-paid leg: a V3_SWAP_EXACT_OUT leg with payerIsUser: true encodes a reduced/omitted PERMIT2_TRANSFER_FROM while keeping the input-token refund SWEEP to spec.recipient, pinning that the direct-leg padding stays with the payer — test/unit/directTransfers.test.ts (in budgeted mode — ingress remainder, ~line 625).

and then some edge cases too we could test for

- Off-token SETTLE_ALL is rejected: a v4 SETTLE_ALL on a currency other than the spec input token throws USER_PAID_INPUT_TOKEN_MISMATCH instead of being silently counted into the budget — test/unit/directTransfers.test.ts (in budgeted mode — inbound, ~line 411).
- v4 OPEN_DELTA credit matches across address casing: a checksummed currency in SWAP_EXACT_IN and a lowercase currency in the sole OPEN_DELTA TAKE still produce the credit — test/unit/directTransfers.test.ts (in the OPEN_DELTA credits describe, ~line 847).
- MAX_UINT160 pull boundary: a user-paid pull of exactly MAX_UINT160 is accepted and MAX_UINT160 + 1 is rejected with USER_PAID_AMOUNT_OUT_OF_RANGE — test/unit/directTransfers.test.ts (next to the zero/sentinel rejection test at ~line 452).
- Zero-amount exact-input omits ingress: an exactOrMaxAmountIn of 0 encodes no PERMIT2_TRANSFER_FROM command at all (the behavior change called out in the PR description) — test/unit/encodeSwaps.test.ts if not already pinned there.
- Default-regime ingress is never reduced: with allowDirectTransfers unset, the encoded ingress always equals the full exactOrMaxAmountIn regardless of step contents, making the encoder/validator coupling explicit — test/unit/directTransfers.test.ts (only needed if the one-line flag gate in swapRouter.ts isn’t taken).

PR #615, opened 2026-06-23
feat(uniswapx-sdk): add Robinhood (4663) + Arc (5042) for DutchV3 rollout

codyborn:
## Summary

Adds **Robinhood (4663)** and **Arc (5042)** to `@uniswap/uniswapx-sdk` so the DutchV3 reactors deployed on both chains ([Uniswap/UniswapX#371](https://github.com/Uniswap/UniswapX/pull/371)) are resolvable. This is the Phase-2 gate in the rollout playbook — the parameterization-api, x-service, and trading-api all pin this release.

This branch already carried the `PERMIT2_MAPPING` additions + the 3.0.8 version bump; this PR completes the wiring (a chain in `PERMIT2_MAPPING` but missing from `REACTOR_ADDRESS_MAPPING` can't actually build an order, so these belong in the same release).

## Changes (`src/constants.ts`)

| Mapping | 4663 / 5042 |
|---|---|
| `REACTOR_ADDRESS_MAPPING[Dutch_V3]` | `0x0000…3cBA` (Robinhood) / `0x0000…2C36` (Arc) |
| `UNISWAPX_ORDER_QUOTER_MAPPING` | canonical `0x00000000a3db…7F58` (explicit — would otherwise fall through to the legacy `0x5453…` default) |
| `EXCLUSIVE_FILLER_VALIDATION_MAPPING` | zero address (explicit — default fall-through is `0x8A66…`, which isn't deployed on either chain; exclusivity is reactor-enforced) |
| `PERMIT2_MAPPING` | already present (prior commit) |

No `Priority` / `Dutch_V2` / `V4` entries — their absence is the upstream guard that makes x-service's `OffChainUniswapXOrderValidator` reject those order types on these chains.

## Verification

- Addresses match the on-chain deployments (`owner()` / `permit2()` checked live; x-contracts V3 integration suite passes 6/6 on each chain)
- New tests: per-chain `getReactor` / quoter / permit2 / exclusive-filler assertions, plus a `V3DutchOrderBuilder` round-trip decaying over a chain-realistic block window (Arc 60 blocks ≈ 30s at 500ms; Robinhood 120 ≈ 30s at 250ms)
- `bun test src/` → 328 pass; `bun run build` (cjs/esm/types) clean

## Test plan
- [x] `bun test src/`
- [x] `bun run build`
- [ ] Publish 3.0.8 and bump the pin in parameterization-api / x-service / trading-api

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #682, opened 2026-08-05
feat(uniswapx-sdk): add Ink (57073) for DutchV3 rollout

codyborn:
Phase 2 of the Ink (chainId 57073) DutchV3 rollout. The reactor and OrderQuoter are deployed and source-verified on Ink (Uniswap/UniswapX#375); this PR adds the mapping entries that let the SDK resolve them.

## Changes

| Mapping | Entry for 57073 |
|---|---|
| `REACTOR_ADDRESS_MAPPING[57073][Dutch_V3]` | `0x000000007A1C8e570011EeDF86A2A35593013cBA` |
| `UNISWAPX_ORDER_QUOTER_MAPPING[57073]` | canonical `0x00000000a3db63Df9078cBF3dF88B4CAdD5a7F58` |
| `PERMIT2_MAPPING[57073]` | canonical Permit2 |
| `EXCLUSIVE_FILLER_VALIDATION_MAPPING[57073]` | zero address |

Ink is not in `NETWORKS_WITH_SAME_ADDRESS`, so all four need explicit entries. `constructSameAddressMap` only seeds those five chains — there is no fallthrough default for an arbitrary chain, so a missing entry reads back `undefined`. The two consequences differ in loudness:

- **Quoter** — `UniswapXOrderQuoter`'s constructor hits the `else` branch and throws `MissingConfiguration("quoter", "57073")`. Loud, but quoting is impossible without the entry.
- **Exclusive-filler validator** — the quiet one. `encodeExclusiveFillerData` assigns the lookup into `ValidationInfo.additionalValidationContract` with no guard, so `undefined` propagates into order construction rather than throwing. Zero address matches every other V3-only chain; exclusivity itself is reactor-enforced via `ExclusivityLib`, so zero makes the hook correctly inert.

Both cases are covered by the tests below. The test-file comment describing this is corrected in the same pass — it previously described both as falling through to deployed-but-wrong defaults, which is not what the map does.

## The reactor address is shared with Robinhood

Not a typo — Ink's reactor is deployed at the same address as Robinhood's. This is already the status quo elsewhere: chains 130, 196, and 1868 share `0x000000005aF6…` today.

`REVERSE_REACTOR_MAPPING` is keyed by reactor address, so it collapses the two chains to a single key. That is correct here since both map to `Dutch_V3`. I verified at runtime that the shared key resolves to `Dutch_V3`, that both chains return the same reactor from `getReactor`, and that no address anywhere in the mapping is claimed by two *different* order types — the property that would actually matter if it broke.

## What is deliberately absent

No `Priority`, `Dutch_V2`, `Hybrid`, or `UNISWAPX_V4_*` entries. No such reactors are deployed on Ink, and the absence is the upstream guard: x-service's `OffChainUniswapXOrderValidator.validateReactorAddress` infers "unsupported order type" from the missing entry. Adding zero-address placeholders would defeat that.

## No chain-specific order-construction handling needed

Ink is a standard OP-stack L2 — 1s contiguous blocks, real dynamic EIP-1559 basefee, native ETH with the canonical WETH predeploy — so `adjustmentPerGweiBaseFee` keeps its normal non-zero treatment and the native sentinel is usable. `@uniswap/sdk-core` already has full Ink coverage (`ChainId.INK`, `INK_ADDRESSES`, `WETH9[57073]`, 1s block time), added in #614, so no sdk-core change is needed.

## Tests

Extends the two existing DutchV3-rollout tables (added for Robinhood/Arc in #615) with Ink rather than adding parallel cases:

- `constants.test.ts` — `getReactor` resolution, canonical quoter, canonical Permit2, zero-address filler validation. Inline `REACTOR_ADDRESS_MAPPING` snapshot updated.
- `V3DutchOrderBuilder.test.ts` — builds a V3 order for 57073 with a **30-block** decay window (~30s wallclock at Ink's 1s blocks; compare 120 for Robinhood, 60 for Arc) and asserts the reactor resolves without an explicit address.

Full `uniswapx-sdk` unit suite passes: **333 tests, 0 failures**.

Note the three touched files are already Prettier-dirty on `main`, so I did not run `prettier --write` — it would have buried this change in unrelated reformatting. The added lines follow the surrounding style.

## Release

Ships as a `patch` via changeset (`.changeset/ink-dutchv3-mappings.md`) rather than a hand-edited `package.json` + `CHANGELOG.md`, so the release bot handles it and this doesn't conflict with the pending changeset on `main`.


.changeset/ink-dutchv3-mappings.md
```diff
@@ -0,0 +1,14 @@
+---
+'@uniswap/uniswapx-sdk': patch
+---
+
+Add Ink (chainId 57073) DutchV3 support. The reactor and OrderQuoter are deployed and source-verified on Ink; these are the mapping entries that let the SDK resolve them.
+
+- `REACTOR_ADDRESS_MAPPING[57073][Dutch_V3]` → `0x000000007A1C8e570011EeDF86A2A35593013cBA`. This is the same address as Robinhood (4663), as chains 130/196/1868 already share `0x000000005aF6…`; `REVERSE_REACTOR_MAPPING` collapses shared addresses to one key, which is correct since both are `Dutch_V3`.
+- `UNISWAPX_ORDER_QUOTER_MAPPING[57073]` → canonical `0x00000000a3db63Df9078cBF3dF88B4CAdD5a7F58`. Required: without it the lookup falls through `constructSameAddressMap` to the legacy `0x54539967…` default, which is not deployed on Ink.
```
ericneil-sanc: constructSameAddressMap only populates the five NETWORKS_WITH_SAME_ADDRESS chains, so without this entry UNISWAPX_ORDER_QUOTER_MAPPING[57073] is undefined and the UniswapXOrderQuoter constructor throws MissingConfiguration — it fails loudly rather than silently resolving to 0x5453

.changeset/ink-dutchv3-mappings.md
```diff
@@ -0,0 +1,14 @@
...
+- `PERMIT2_MAPPING[57073]` → canonical Permit2, verified on-chain (9,152 bytes, non-zero `DOMAIN_SEPARATOR()`).
+- `EXCLUSIVE_FILLER_VALIDATION_MAPPING[57073]` → zero address, matching every other V3-only chain. Required: the default is `0x8A66…`, a contract that does not exist on Ink. Exclusivity is reactor-enforced via `ExclusivityLib`.
```
ericneil-sanc: the lookup yields undefined, not 0x8A66… (the silent bug there is encodeExclusiveFillerData passing undefined through into ValidationInfo).

ericneil-sanc (approved):

.changeset/ink-dutchv3-mappings.md
codyborn: You're right, and I checked before changing it — `constructSameAddressMap` only seeds the five `NETWORKS_WITH_SAME_ADDRESS` chains, so there's no fallthrough default at all. Confirmed empirically against an absent chain: `UNISWAPX_ORDER_QUOTER_MAPPING[59144]` is `undefined`, and only chain `5` is still on the legacy `0x5453…` (the other four seeded chains are overridden by explicit entries below).

So the failure is the `else` branch in `UniswapXOrderQuoter`'s constructor throwing `MissingConfiguration("quoter", "57073")` — loud, not silent. Corrected in the changeset, the PR body, and the test-file comment.

That comment predates this PR (it came in with the Robinhood/Arc rollout) and I propagated it when extending the table, so the fix covers the original wording too.

.changeset/ink-dutchv3-mappings.md
codyborn: Correct on both points — the lookup is `undefined`, not `0x8A66…`, and the real hazard is downstream: `encodeExclusiveFillerData` does

```ts
} else if (chainId) {
  additionalValidationContract = EXCLUSIVE_FILLER_VALIDATION_MAPPING[chainId];
}
```

with no guard, so `undefined` lands in the returned `ValidationInfo` rather than throwing. That makes this the quiet one of the two and the entry more load-bearing than I'd written, not less. Rewritten accordingly.

PR #705, opened 2026-08-24
fix(uniswapx-sdk): resolve V3 Dutch orders exactly as the reactor settles them

alanhwu:
## Problem

`CosignedV3DutchOrder.resolve()` decayed the raw curve and ignored the rest of the signed order, so it could report amounts that differ from — and in the worst case invert — the ones `V3DutchOrderReactor` settles with. `BaseReactor.execute()` transfers the reactor's amounts, so a filler that prices an order with `resolve()` and then fills it pays what the reactor resolved, not what it was shown.

With `input {start: 1, max: 1, curve: -999_999}` and `output {start: 1_000_000, min: 1_000_000, curve: +999_999}`, the old resolver reported an input of 1,000,000 and an output of 1, where the reactor settles an input of 1 and an output of 1,000,000.

## Fix

`resolve()` is now a faithful port of `_resolve`: cosigner overrides → base fee adjustment → bounded decay → exclusivity override.

| Divergence | Reactor | Before |
|---|---|---|
| Bounds | input to `[0, maxAmount]`, output to `[minAmount, uint256.max]` | neither applied |
| Base fee adjustment | `_updateWithGasAdjustment` before decay | `startingBaseFee` and both `adjustmentPerGweiBaseFee` dropped |
| Rounding | separate input/output funcs on *relative* amounts, opposite directions | one shared round-down on *absolute* amounts |
| Exclusivity | outputs scaled by `exclusivityOverrideBps`; strict exclusivity rejects | not applied |
| Cosigner amounts | reverts when an override worsens the order | accepted |

`decay()` also caps `blockDelta` at `uint16` max, as the reactor does.

The rounding branch was inverted, not just imprecise: interpolating absolute amounts flips floor/ceil relative to interpolating the relative amounts. It was off by one against the reactor for inputs on a decaying curve and outputs on an increasing curve — in both cases reading in the filler's favor and settling against it.

`MathExt.sol` is ported to `utils/mathExt.ts` and `ExclusivityLib.sol` to `utils/exclusivity.ts`. The exclusivity helper is generic over the output shape and over position (timestamp or block), so V1 and V2 — which have the same omission — can adopt it without an import cycle. `getBlockDecayedAmount` is replaced by `decayInput`/`decayOutput`, which take the signed input/output so the bounds cannot be dropped at the call site.

## Breaking changes

The public API is unchanged: `package.json` exposes only `"."` and `utils/index.ts` never re-exported `./dutchBlockDecay`, so the replaced symbols were unreachable from the published package. `V3OrderResolutionOptions` gains only optional fields.

The breaks are behavioral, ranked by how likely they are to fire:

1. **Strict exclusivity now throws.** `resolve({ currentBlock })` with no `filler`, on an order with a real `exclusiveFiller` and `exclusivityOverrideBps === 0`, throws `NoExclusiveOverride` where it previously returned amounts the reactor would never settle. `V3DutchOrderBuilder` defaults the bps to `0` (`STRICT_EXCLUSIVITY`), so any flow that sets an exclusive filler and leaves the default produces this shape. **Worth confirming against what the parameterization API actually cosigns before release** — callers fix it by passing `options.filler`, but an unguarded order loop would throw rather than skip one order.
2. **Resolved amounts change.** Correct now, but different; anything snapshotting or caching `resolve()` output will move. For well-formed orders with zero gas adjustment and non-crossing curves the only delta is the ±1 wei rounding.
3. **Throws on nonzero `adjustmentPerGweiBaseFee` without `blockBaseFee`**, rather than silently returning an unadjusted amount.
4. **Throws on invalid cosigner amounts** — only for orders the reactor rejects anyway.

Changeset is marked `minor`. Under strict semver, `resolve()` throwing where it previously returned is a major — flagging for a deliberate call. If the exclusivity change is the sticking point, it can be split out so the bounds and base-fee fixes ship without callers auditing their call sites first.

## Testing

376 unit tests pass, lint and build clean. New coverage: bound clamping in both directions, base fee adjustment on an empty curve, input/output rounding divergence, `uint16` block-delta cap, exclusivity scaling and strict-exclusivity rejection, and all three cosigner validation paths.

`V3DutchOrder.ts` was already prettier-nonconforming on `main` (4-space indent), so it keeps the file's existing style rather than burying the diff in a reflow.


sdks/uniswapx-sdk/src/order/V3DutchOrder.test.ts
```diff
@@ -270,5 +270,232 @@ describe("V3DutchOrder", () => {
             expect(resolved.outputs[0].token).eq(order.info.outputs[0].token);
             expect(resolved.outputs[0].amount).eq(order.info.cosignerData.outputOverrides[0]);
         });
+
+        // The reactor bounds the decayed input to the signed maxAmount and the
+        // decayed output to the signed minAmount. Resolving without those
+        // bounds reports economics inverted from the ones the reactor settles
+        // with, which would drain a filler that priced the order locally.
+        it("bounds a curve that resolves across maxAmount and minAmount", () => {
```
gretzke, line 278: These assert values copied in by hand rather than values the reactor produced. That is the same condition that let the original divergence exist, so the suite cannot demonstrate the property in the title.

What would pin it is a differential test that generates V3 orders and asserts `resolve()` equals the on-chain `OrderQuoter` across a range of blocks, base fees and fillers. The quoter runs the real reactor and returns the resolved order through the callback revert, so it is an independent oracle rather than a second copy of the same assumption.

This matters outside the repo too. Cantina asked on finding #843 whether remediation makes every hard-quote-created V3 order resolve identically in the published SDK and the reactor. Without an oracle test the strongest answer I can give them is a code read.

sdks/uniswapx-sdk/src/order/V3DutchOrder.ts
```diff
@@ -581,42 +592,180 @@ export class CosignedV3DutchOrder extends UnsignedV3DutchOrder {
...
+     * V3DutchOrderReactor._resolve: cosigner overrides are applied to the base
+     * amounts, then the base fee adjustment, then the bounded decay curve, then
+     * the exclusivity override.
+     *
+     * @param options resolution context. `blockBaseFee` is required when the
+     * order carries a nonzero adjustmentPerGweiBaseFee. `filler` is required to
+     * claim exclusive filling rights; without it the order resolves for an
+     * arbitrary filler.
+     * @return the input and outputs the reactor would transfer
+     * @throws when the reactor would reject the order outright, rather than
```
gretzke: This says it throws when the reactor would reject the order outright, but `resolve()` checks neither the deadline nor the cosignature, and the reactor checks both before it touches any amounts. An expired order with a garbage cosignature returns amounts here and reverts on chain.

I would soften the comment rather than add the checks. `resolve()` is a pricing function and validity belongs elsewhere. As written the wording promises more than the code does.

sdks/uniswapx-sdk/src/utils/mathExt.ts
```diff
@@ -0,0 +1,86 @@
...
+  a: BigNumber,
+  b: BigNumber,
+  min: BigNumber,
+  max: BigNumber
+): BigNumber {
+  return boundedSub(a, b.mul(-1), min, max);
+}
+
+/** Mirrors solmate's FixedPointMathLib.mulDivDown. */
+export function mulDivDown(
```
gretzke, line 94: `mulDivDown` and `mulDivUp` multiply at arbitrary precision, where solmate reverts once `x * y` exceeds uint256. The same gap exists in the exclusivity scaling and in `gasDeltaWei`, where the base fee difference can go negative and `SafeCast.toInt256` would revert.

Low severity, because the reactor reverts atomically in all of these cases, so there is no wrong settlement. The cost is a filler pricing an order that looks fillable and then burning gas on a transaction that can only revert. A note in the code seems better than reproducing Solidity overflow semantics in TypeScript.

sdks/uniswapx-sdk/src/utils/dutchBlockDecay.ts
```diff
@@ -1,124 +1,250 @@
...
+/**
+ * Locates the curve segment containing `currentRelativeBlock` and returns its
+ * bounding points along with their amounts relative to the start amount.
+ * Mirrors NonlinearDutchDecayLib.locateCurvePosition.
+ */
+function locateCurvePosition(
+  curve: NonlinearDutchDecay,
+  currentRelativeBlock: number
+): [number, number, BigNumber, BigNumber] {
+  const { relativeBlocks, relativeAmounts } = curve;
```
gretzke, line 66: Two pre-existing problems in the packed representation these values come from, neither introduced here, but worth recording while this is open.

`encodeRelativeBlocks` ors each value in at `i * 16` with no uint16 mask, so anything above 65535 writes into the next slot. `[65536, 5]` packs to a value that both the reactor and `decodeRelativeBlocks` read back as `[0, 5]`, replacing a curve point without error. `V3DutchOrderBuilder` checks ordering but not range, so nothing rejects it.

`decodeRelativeBlocks` also calls `.toNumber()` on the whole packed value before masking, which throws `NUMERIC_FAULT` once the packed value passes the JS safe integer range. That happens at five curve points, where the reactor supports sixteen, so multi-point curves cannot be parsed at all.

I checked whether the first one can mislead a filler through the order feed and it cannot. `decodeRelativeBlocks` masks with `0xffff`, matching `Uint16ArrayLibrary.getElement`, so anything arriving as `encodedOrder` bytes resolves identically on both sides. The gap only exists for an order object built in memory and never round tripped.

.changeset/v3-resolve-matches-reactor.md
```diff
@@ -0,0 +1,22 @@
+---
+'@uniswap/uniswapx-sdk': minor
```
gretzke: Registering a view on the question you raised. `resolve()` throwing where it previously returned is a major under strict semver, and three of the four behavioral breaks are new throws.

Splitting exclusivity out is not needed for release safety though. The parameterization API never cosigns a strictly exclusive V3 order. The RFQ path sets `exclusivityOverrideBps` to 25, and the open order path sets `exclusiveFiller` to the zero address, which `hasFillingRights` short circuits. So nothing published today can reach the `NoExclusiveOverride` throw.

gretzke (approved):

PR #706, opened 2026-08-25
fix(router-sdk): fix MixedRouteSDK midPrice across native/wrapped boundaries (ROUTE-886)

gretzke:
`MixedRouteSDK.midPrice` selects the direction of each pool price with an exact currency equality check. When a route crosses a native/wrapped boundary, for example a v2 WETH pair followed by a v4 pool that holds native ETH, the exact check fails. The reducer then multiplies by the inverted pool price.

This caused a production incident on Robinhood Chain. A MIXED route went token to WETH on a v2 pair, then through a native ETH v4 pool. The mid price came out 13 orders of magnitude too small. The computed price impact was about -1e14 percent instead of +90 percent. The consumer clamped the value to -100 percent and the price impact warning never fired. Internal reference: ROUTE-886.

The constructor's tokenPath loop already resolves this boundary. The `midPrice` getter now matches it. The reducer checks exact equality against token0 and token1 first, then falls back to wrapped equality. If nothing matches, it throws `PRICE_PATH` instead of multiplying a wrong price with no error. The exact checks stay first so a genuine ETH/WETH pool keeps an unambiguous direction.

Tests added:

- Two regression tests that pin the inversion in both directions. Against the old code they produce 4.0000 instead of 0.2500 and 1.3333 instead of 12.0000.
- Two tests with a genuine (non-fake) ETH/WETH v4 pool at an asymmetric price. Both sides of that pool wrap to WETH, so these pin the exact-before-wrapped ordering.

Changeset: patch bump for @uniswap/router-sdk.

### DIFF
--- a/sdks/router-sdk/src/utils/index.ts
+++ b/sdks/router-sdk/src/utils/index.ts
@@ -5,6 +5,9 @@ import { Pool as V4Pool } from '@uniswap/v4-sdk'
 import { MixedRouteSDK } from '../entities/mixedRoute/route'
 import { TPool } from './TPool'
 
+// `Pair.involvesToken` and `V3Pool.involvesToken` only accept a `Token`, so calling `involvesToken`
+// on the `TPool` union with a bare `Currency` (e.g. native ETH from `route.path`) fails to
+// typecheck. This checks the same token0/token1 membership but against any `Currency`.
 const poolInvolvesCurrency = (pool: TPool, currency: Currency): boolean =>
   pool.token0.equals(currency) || pool.token1.equals(currency)
 
--- a/sdks/router-sdk/src/utils/index.ts
+++ b/sdks/router-sdk/src/utils/index.ts
@@ -1,9 +1,12 @@
-import { Currency, Token } from '@uniswap/sdk-core'
+import { Currency } from '@uniswap/sdk-core'
 import { Pair } from '@uniswap/v2-sdk'
 import { Pool as V3Pool } from '@uniswap/v3-sdk'
 import { Pool as V4Pool } from '@uniswap/v4-sdk'
 import { MixedRouteSDK } from '../entities/mixedRoute/route'
 import { TPool } from './TPool'
+
+const poolInvolvesCurrency = (pool: TPool, currency: Currency): boolean =>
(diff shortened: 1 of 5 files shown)


codyborn (approved):

sdks/router-sdk/src/utils/index.ts
```diff
@@ -43,11 +49,13 @@ export const partitionMixedRouteByProtocol = (route: MixedRouteSDK<Currency, Cur
 export const getOutputOfPools = (pools: TPool[], firstInputToken: Currency): Currency => {
   const { inputToken: outputToken } = pools.reduce(
     ({ inputToken }, pool: TPool): { inputToken: Currency } => {
-      if (!pool.involvesToken(inputToken as Token)) throw new Error('PATH')
-      const outputToken: Currency = pool.token0.equals(inputToken) ? pool.token1 : pool.token0
-      return {
-        inputToken: outputToken,
-      }
+      // exact matches take priority so genuine ETH/WETH pools resolve to the correct side;
```
dannythedawger, line 55: it seems like we didn't have unit tests for this logic in the past. given the importance, I wonder if it makes sense to add tests firsts, make sure they pass. then add the changes and make sure they still pass.

sdks/router-sdk/src/utils/index.ts
```diff
@@ -1,10 +1,13 @@
-import { Currency, Token } from '@uniswap/sdk-core'
+import { Currency } from '@uniswap/sdk-core'
 import { Pair } from '@uniswap/v2-sdk'
 import { Pool as V3Pool } from '@uniswap/v3-sdk'
 import { Pool as V4Pool } from '@uniswap/v4-sdk'
 import { MixedRouteSDK } from '../entities/mixedRoute/route'
 import { TPool } from './TPool'
 
+const poolInvolvesCurrency = (pool: TPool, currency: Currency): boolean =>
```
dannythedawger, line 11: I wonder how different this is from `involvesToken`

dannythedawger (approved):
lgtm. left some ideas. I wonder if it makes sense to also add some higher level unit tests that invoke this code

PR #708, opened 2026-08-28
feat(universal-router-sdk): routerBalanceInput for swaps funded from the router's balance

david-uniswap:
## Description

Adds a `SwapOptions.routerBalanceInput` option that funds a swap from the Universal Router's **own balance** of the input token, spending whatever it holds at execution time instead of pulling a fixed amount from a payer.

The motivating flow is a bridge filler that deposits into the router and swaps atomically in the same transaction: the delivered amount isn't known when the calldata is built, so the first hop has to resolve at execution rather than encode a quoted number.

`useRouterBalance` already exists and sets `payerIsUser = false`, but it keeps the fixed quoted `amountIn`, so it means "the router holds a known amount". This is the other case.

## What it does

- First hop encodes `CONTRACT_BALANCE` and `payerIsUser = false`. Later hops already chain through `CONTRACT_BALANCE`, so only hop 0 changes.
- Optional `minimumAmount` emits a `BALANCE_CHECK_ERC20` before any swap, so an under-funded router reverts up front. This is a **different guarantee from `slippageTolerance`**: the trade-level minimum output only catches a shortfall large enough to breach it, so a wide tolerance can let an under-delivery through. Bounding the input directly closes that.
- The trade-level minimum output is unchanged, so price protection still comes from the SDK rather than the caller.

```ts
SwapRouter.swapCallParameters(trade, {
  recipient,
  slippageTolerance,
  chainId,
  routerBalanceInput: { minimumAmount: '999000000' }, // minimumAmount optional
})
```

## Guards

All of these throw in the `UniswapTrade` constructor rather than silently encoding something wrong:

- **Explicit `recipient` required.** `SENDER_AS_RECIPIENT` resolves to `msg.sender`, who in this flow is the funder, not the beneficiary. Silently paying the filler is the worst failure mode here, and it mirrors the existing `ApproveProxy` guard.
- **ERC20 input**, **`EXACT_INPUT`**, and a **single non-split route**. One `CONTRACT_BALANCE` cannot address two legs of the same currency: the first drains it and the second resolves to zero.
- Incompatible with `inputTokenPermit`, `nativeErc20Input`, and `ApproveProxy`.
- `minimumAmount` requires `chainId`, because `BALANCE_CHECK_ERC20` reads its `owner` param **verbatim** (no sentinel resolution), so it needs the router's real address. Passing `ROUTER_AS_RECIPIENT` there would check the balance of `0x…02`.

## Route coverage

v2, v3, v4 and mixed routes are all supported.

Pure v4 needed one extra step: `V4Planner.addTrade` bakes the quoted `amountIn` into the swap action, so for this mode the encoder builds the `SETTLE(CONTRACT_BALANCE)` + `SWAP_EXACT_IN(open delta)` pair explicitly instead. That's the same shape the mixed-route encoder already uses for its v4 sections, so no v4-sdk change was needed. The settle deliberately precedes the swap, since the swap consumes the delta the settle opens.

## How Has This Been Tested?

`test/unit/routerBalanceInput.test.ts`, 16 cases: every guard above, plus decoded-calldata assertions that the first hop is `CONTRACT_BALANCE` with `payerIsUser = false` on v2 and v3, that the quoted amount is untouched when the option is absent, that the trade-level minimum output is still enforced, and that `BALANCE_CHECK_ERC20` is emitted against the resolved router address with the right token and floor, and that a pure v4 route settles `CONTRACT_BALANCE` before swapping the open delta.

Full unit suite: 492 passing. The one failing test is a pre-existing fork-dependent `before all` hook (476 passing / 1 failing on `main` without this change).

No. New optional field; existing behaviour is unchanged when it's absent, which the "keeps the quoted amountIn" test pins.

## (Optional) Feedback Focus

The option shape. I made it an object so a minimum can't be expressed without the mode, but a flat `routerBalanceInput?: boolean` plus a separate field would be more consistent with the other flat `SwapOptions` entries. Happy either way.

### DIFF
--- a/sdks/universal-router-sdk/src/utils/routerBalanceSteps.ts
+++ b/sdks/universal-router-sdk/src/utils/routerBalanceSteps.ts
@@ -1,0 +1,96 @@
+import invariant from 'tiny-invariant'
+import { CONTRACT_BALANCE } from './constants'
+import { SwapStep, V4Action } from '../types/encodeSwaps'
+
+function v4ActionSpendsToken(action: V4Action, tokenAddress: string): boolean {
+  switch (action.action) {
+    case 'SETTLE':
+    case 'SETTLE_ALL':
+      return action.currency.toLowerCase() === tokenAddress
+    case 'SWAP_EXACT_IN':
+      return action.currencyIn.toLowerCase() === tokenAddress
+    case 'SWAP_EXACT_IN_SINGLE': {
+      const spent = action.zeroForOne ? action.poolKey.currency0 : action.poolKey.currency1
+      return spent.toLowerCase() === tokenAddress
+    }
+    default:
+      return false
+  }
+}
+
+// Whether a step draws the trade's input token, i.e. is a candidate first hop.
+export function stepSpendsToken(step: SwapStep, inputTokenAddress: string): boolean {
+  const tokenAddress = inputTokenAddress.toLowerCase()
+  switch (step.type) {
+    case 'V2_SWAP_EXACT_IN':
+      return step.path[0]?.toLowerCase() === tokenAddress
+    case 'V3_SWAP_EXACT_IN':
+      // v3 exact-in paths are encoded input-first: the first 20 bytes are the input token
+      return step.path.slice(0, 42).toLowerCase() === tokenAddress
+    case 'V4_SWAP':
+      return step.v4Actions.some((action) => v4ActionSpendsToken(action, tokenAddress))
+    default:
+      return false
+  }
+}
+
+function applyToV4Actions(actions: V4Action[], tokenAddress: string): V4Action[] {
+  const hasInputSettle = actions.some(
+    (action) => action.action === 'SETTLE' && action.currency.toLowerCase() === tokenAddress
+  )
+
+  const transformed: V4Action[] = actions.map((action) => {
+    // The settle that funds the swap now takes the router's whole balance.
+    if (action.action === 'SETTLE' && action.currency.toLowerCase() === tokenAddress) {
+      return { ...action, amount: CONTRACT_BALANCE.toString(), payerIsUser: false }
+    }
+    // With the settle sized by CONTRACT_BALANCE, the swap consumes the open delta.
+    if (action.action === 'SWAP_EXACT_IN' && action.currencyIn.toLowerCase() === tokenAddress) {
+      return { ...action, amountIn: 0 }
+    }
+    if (action.action === 'SWAP_EXACT_IN_SINGLE' && v4ActionSpendsToken(action, tokenAddress)) {
+      return { ...action, amountIn: 0 }
+    }
+    return action
+  })
+
+  if (hasInputSettle) {
+    return transformed
+  }
+  // No settle in the plan (addTrade-style shapes): fund the open delta explicitly.
+  return [
+    { action: 'SETTLE', currency: tokenAddress, amount: CONTRACT_BALANCE.toString(), payerIsUser: false },
+    ...transformed,
+  ]
+}
+
+/**
+ * Rewrites the first hop of a step plan to spend the router's entire input-token balance:
+ * v2/v3 exact-in amounts become the CONTRACT_BALANCE sentinel; a v4 first step settles
+ * CONTRACT_BALANCE and swaps the resulting open delta. Later hops already chain through
+ * CONTRACT_BALANCE / open deltas, so only the input-spending step changes.
+ *
+ * `validateEncodeSwaps` guarantees exactly one step spends the input token and that it is
+ * the first step, so this only ever rewrites `steps[0]`.
+ */
+export function applyRouterBalanceInputToSteps(swapSteps: SwapStep[], inputTokenAddress: string): SwapStep[] {
+  const tokenAddress = inputTokenAddress.toLowerCase()
+  const first = swapSteps[0]
+  invariant(first !== undefined && stepSpendsToken(first, tokenAddress), 'ROUTER_BALANCE_INPUT_FIRST_STEP')
+
+  let transformed: SwapStep
+  switch (first.type) {
+    case 'V2_SWAP_EXACT_IN':
+    case 'V3_SWAP_EXACT_IN':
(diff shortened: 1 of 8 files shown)


sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
```diff
@@ -132,6 +133,41 @@ export function validateEncodeSwaps(spec: NormalizedSwapSpecification, swapSteps
...
+    const spenderIndexes = swapSteps
+      .map((step, index) => (stepSpendsToken(step, balanceInputTokenAddress) ? index : -1))
+      .filter((index) => index >= 0)
+    invariant(spenderIndexes.length === 1 && spenderIndexes[0] === 0, 'ROUTER_BALANCE_INPUT_SPLIT_ROUTE')
+    for (const step of swapSteps) {
+      invariant(
+        step.type !== 'V2_SWAP_EXACT_OUT' && step.type !== 'V3_SWAP_EXACT_OUT',
+        'ROUTER_BALANCE_INPUT_EXACT_INPUT_ONLY'
+      )
+      invariant(step.type !== 'WRAP_ETH', 'ROUTER_BALANCE_INPUT_NATIVE_INPUT')
```
scorzeth, line 167: Its possible for wraps to happen mid-route, when a v4 leg pays out native ETH and a later leg or the output needs WETH, a router will insert a WRAP_ETH partway through the plan, already sized CONTRACT_BALANCE.

sdks/universal-router-sdk/src/utils/routerBalanceSteps.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Rewrites the first hop of a step plan to spend the router's entire input-token balance:
+ * v2/v3 exact-in amounts become the CONTRACT_BALANCE sentinel; a v4 first step settles
+ * CONTRACT_BALANCE and swaps the resulting open delta. Later hops already chain through
+ * CONTRACT_BALANCE / open deltas, so only the input-spending step changes.
+ *
+ * `validateEncodeSwaps` guarantees exactly one step spends the input token and that it is
+ * the first step, so this only ever rewrites `steps[0]`.
+ */
+export function applyRouterBalanceInputToSteps(swapSteps: SwapStep[], inputTokenAddress: string): SwapStep[] {
```
scorzeth, line 76: I'm worried that having the SDK rewrite steps could get a bit hairy, maybe instead we should make the routers produce the correct steps for this situation instead?

sdks/universal-router-sdk/src/utils/routerBalanceSteps.ts
```diff
@@ -0,0 +1,96 @@
...
+  )
+
+  const transformed: V4Action[] = actions.map((action) => {
+    // The settle that funds the swap now takes the router's whole balance.
+    if (action.action === 'SETTLE' && action.currency.toLowerCase() === tokenAddress) {
+      return { ...action, amount: CONTRACT_BALANCE.toString(), payerIsUser: false }
+    }
+    // With the settle sized by CONTRACT_BALANCE, the swap consumes the open delta.
+    if (action.action === 'SWAP_EXACT_IN' && action.currencyIn.toLowerCase() === tokenAddress) {
+      return { ...action, amountIn: 0 }
```
scorzeth, line 49: This sets amountIn to 0 on every swap that spends the input token. That works when the step has one such swap. But the router's usual v4 first hop is one SETTLE that deposits the input, followed by several swaps that each take a fixed slice of it. After this rewrite the first swap takes the whole deposit and the second swap has nothing left. A zero-amount swap is rejected by the pool manager, and that swap's own minimum output could not be met anyway, so the transaction reverts.

scorzeth (commented):
Overall I'm concerned with the rewriting of the SwapSteps plans (in this and #710). The intention of this move (to encodeSwaps/SwapSteps) was to allow the routers to control their steps, and for the SDK to handle the safety envelope around it:  ingress, permits, fees, the sweep and its floors, and validation.

sdks/universal-router-sdk/src/utils/validateEncodeSwaps.ts
david-uniswap: Yep, fixed. WRAP_ETH is only refused at hop 0 in ERC20 balance mode now. Mid-route wraps pass through untouched.

sdks/universal-router-sdk/src/utils/routerBalanceSteps.ts
david-uniswap: Fair worry, and the payerIsUser bug proves the rewrite is fragile. But pushing it into routers means GuideStar, UniRoute and Quickroute each reimplement balance mode and we end up with three versions of the same edge cases. Fixed the hairy bits instead: every spender leg gets payerIsUser cleared, v4 splits are supported, the largest leg goes last, and a v4 plan that swaps before settling is now refused. Tests for each.

sdks/universal-router-sdk/src/utils/routerBalanceSteps.ts
david-uniswap: Good catch, that was wrong. Now only the largest input swap becomes open-delta, and it's moved last so the fixed slices run first. Covered in encodeSwaps.test.ts (multi-swap v4 hop case).

david-uniswap:
On the bigger point about routers controlling their steps: agree on the split, and I'd argue balance mode is part of the envelope, not the route. It's ingress: who pays and where the input comes from. Cody's take is the same: routers should only know what they need to produce a path, and TAPI owns execution mode so it stays unified across routers. So v1 keeps the rewrite in the SDK (called from TAPI), and long term we want routers emitting a plainer route shape with no payer or custody fields at all. Would love your help spec'ing that.

Also addressed from the TDD thread: v4 in splits is supported (both shapes tested), payerIsUser is forced false on every spender leg, and native ETH into a raw-native v4 leg is refused with a typed error since native mode needs a wrap step to check the balance floor as WETH.

scorzeth (approved):
should go in at the same time as #710, approving

PR #710, opened 2026-08-31
feat(universal-router-sdk): native input for routerBalanceInput via msg.value + full wrap

david-uniswap:
## Description

Follow-up to #708. Adds native-ETH funding to `routerBalanceInput`, since ETH is the most common bridged asset.

Raw ETH transfers to the Universal Router revert (`receive()` only accepts ETH from WETH), so the deposit-then-execute pattern can't fund a native balance swap. Instead the executor attaches the delivered amount as `msg.value` on `execute()` and the plan:

1. `WRAP_ETH(router, CONTRACT_BALANCE)` — wraps everything the router holds, so no value is left behind (UR never refunds `msg.value`)
2. optional `BALANCE_CHECK_ERC20` on WETH **after** the wrap (there is no native balance-check command)
3. swaps the wrapped balance (hop 0 stays `CONTRACT_BALANCE`)
4. always appends a trailing `SWEEP(ETH, recipient, 0)` so any ETH dust goes to the recipient — the funder is a third party, so leftover value would otherwise be stranded

The encoded `MethodParameters.value` is `0`: the delivered amount is unknown at encode time and the executor supplies it. Pure-native v4 routes (no wrap step to size with `CONTRACT_BALANCE`) remain refused with a typed error.

## Tests

- refusal: pure-native v4 route
- accept: native input on a wrapping route
- encode: `WRAP_ETH(CONTRACT_BALANCE)` + `CONTRACT_BALANCE` hop 0 + trailing ETH dust sweep, encoded value 0
- encode: WETH floor asserted post-wrap when `minimumAmount` is set

### DIFF
--- a/.changeset/router-balance-input-native.md
+++ b/.changeset/router-balance-input-native.md
@@ -1,0 +1,7 @@
+---
+'@uniswap/universal-router-sdk': minor
+---
+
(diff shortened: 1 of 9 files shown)


scorzeth (approved):
lgtm!

PR #717, opened 2026-09-01
feat(liquidity-launcher-sdk): point Arc (5042) Instant Launch at the 2026-09-01 redeploy

ciyengar3:
## PR Scope

`feat(liquidity-launcher-sdk)` — minor version bump (→ 1.13.0 when the changesets "Version Packages" PR lands).

## Description

**Before**: the SDK's Arc Instant Launch generation (registered this morning in #713) points at the initial strategy pair, which opened pools at initialTick 198,050.

**After**: it points at the 2026-09-01 same-day strategy redeploy, which opens pools at **initialTick 122,050** (a constructor immutable — the reason for the redeploy). TICK_SPACING 25 and MIN_LAUNCH_TICK −160,100 are unchanged, as are the FeeSplitters, UERC20BeneficiaryVault, CompoundingClaimRecipient and launcher pin.

### How

- `sdks/liquidity-launcher-sdk/src/addresses.ts`
  - `INSTANT_LAUNCH_STRATEGY_FEES_ON_ARC` → `0xfe7Be4EbBE6CcDfA57EE8c36fe9a767B033eB056`, `INSTANT_LAUNCH_STRATEGY_FEES_OFF_ARC` → `0xff301aCB22816D210d75D71F31Ac13C771093EF3`; the two Arc registry entries move to `initialTick: 122050`.
  - Verified on-chain before registering: `beneficiaryVault()` identifies the variants (Arc vault vs `address(0)`), `feeSplitter()` confirms both reuse the existing splitters (splits re-read via `getSplits()`: 40/60+100 and 100/100), `launcher()` is the Arc launcher, and the pool-shape getters return 25 / 122,050 / −160,100.
  - **The superseded pair is replaced, not appended**: it was never launched against — zero logs emitted by either strategy across its entire lifetime (verified via `eth_getLogs` over the full deploy window) — so there are no indexed launches for the append-only rule to protect. Same reasoning as the chain-4663 v3.1.0 dev-pair removal documented in the file.
- Tests (`addresses.test.ts`): the Arc literals move to the new strategies and initialTick 122,050 (independent string literals, not registry-derived).
- `.changeset/arc-instant-launch-redeploy.md`: `minor` bump.

**Intentionally unchanged**: launcher stack (`LAUNCHER_ADDRESSES[5042]`), FeeSplitters, singletons, every other chain. No ABI, type, or function-signature changes.

## How Has This Been Tested?

- `bun test` (package): 238 pass / 0 fail
- `bun run lint`: clean

No — address-registry values only.

## (Optional) Feedback Focus

- Replace-vs-append for the superseded pair: I verified zero on-chain logs from both old strategies before removing them; if anything launched against them between my check and merge, they must be re-added as a historical generation instead.

### DIFF
--- a/.changeset/arc-instant-launch-redeploy.md
+++ b/.changeset/arc-instant-launch-redeploy.md
@@ -1,0 +1,5 @@
+---
+'@uniswap/liquidity-launcher-sdk': minor
+---
+
+Point the Arc (5042) Instant Launch generation at the 2026-09-01 strategy redeploy: fees-on `0xfe7Be4Eb…`, fees-off `0xff301aCB…`, opening pools at initialTick 122,050 (previously 198,050; TICK_SPACING 25 and MIN_LAUNCH_TICK -160,100 unchanged). The superseded initial pair is replaced rather than retained — it was never launched against (zero on-chain logs), so no indexed launches reference it. FeeSplitters, vault and compounding recipient are unchanged.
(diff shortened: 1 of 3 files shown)


ciyengar3:
Superseded by #718, which shipped the same redeploy plus the Arc block time and uERC20 factory.

PR #736, opened 2026-09-18
feat(universal-router-sdk): add Universal Router v2.1.2 addresses

dianakocsis:
## Description

Records the Universal Router **v2.1.2** deployment in the SDK, adding `UniversalRouterVersion.V2_1_2` and the matching `URVersion.V2_1_2`.

Addresses are taken verbatim from [Uniswap/universal-router#516](https://github.com/Uniswap/universal-router/pull/516), covering all **24 chains** that already ran v2.1.1:

mainnet (1), optimism (10), bsc (56), unichain (130), polygon (137), monad (143), xlayer (196), worldchain (480), unichain-sepolia (1301), soneium (1868), tempo (4217), megaeth (4326), robinhood (4663), arc (5042), base (8453), arbitrum (42161), celo (42220), avalanche (43114), ink (57073), linea (59144), blast (81457), base-sepolia (84532), zora (7777777), sepolia (11155111)

Both enums gain the member because `toV4URVersion` resolves across the two by string value, so they have to stay in sync. `isAtLeastV2_1_1('2.1.2')` is `true`, so v2.1.2 inherits the 1e18-precision fee encoding of v2.1.1 with no code change.

Two things worth a reviewer's eye:

- **`V2_2_0` is intentionally left in place.** Despite the higher version string, 2.2.0 is an earlier, separate permissioned-pool deployment (mainnet block `25733844` vs v2.1.2's `25999984`) and is not superseded by this change. The new member is ordered between `V2_1_1` and `V2_2_0`.
- **Ink gets a standalone v2.1.2 router** (`0x661E93cc…`), rather than the v2.2.0 alias that currently backs `V2_1_1` there. The existing alias and its comment are untouched.

### On the creation blocks

PR 516 records addresses but **not** creation blocks, which `RouterConfig` requires. Each block was resolved from chain state rather than guessed:

- **Etherscan V2** `getcontractcreation` for the 15 chains it supports on this key
- **`eth_getCode` bisection** against an archive RPC for the other 9 (optimism, bsc, base, avalanche, xlayer, soneium, tempo, ink, zora), each confirmed as code-at-`N` / no-code-at-`N-1`

Cross-checked two independent ways, both of which passed for all 24:

1. every v2.1.2 block is strictly after that chain's v2.1.1 block
2. all 24 block timestamps fall inside the same ~90 minute window (2026-09-17 21:33–23:00 UTC), consistent with one coordinated multichain deploy — a pruned-node false positive would not cluster with the Etherscan-derived answers

## How Has This Been Tested?

- `universal-router-sdk`: **657 passing** (up from 652), `forge test` 80 passed, clean build, prettier clean
- `v4-sdk`: **240 pass / 2 skip / 0 fail**, clean build, eslint clean
- All 24 addresses diffed byte-for-byte against PR 516 — exact match
- All 24 verified as valid EIP-55 checksummed via `cast to-check-sum-address`

New coverage in `constants.test.ts` pins the invariants: the 2.1.1 and 2.1.2 chain sets are identical, every 2.1.2 block is after its 2.1.1 block, and Ink's 2.1.2 is asserted *not* to alias 2.2.0.

No. The diff is **164 insertions, 0 deletions** — purely additive. New enum members and new `routerConfigs` entries; no existing address, block, type, or signature is touched.

## (Optional) Follow Ups

- Upstream **PR 516 is still open**. If any address changes before it merges, this PR needs a matching update — worth confirming before release.
- The parked `remove-ur-v2-2-0` work is unaffected and still applies cleanly on top of this if you want to retire 2.2.0 separately.


codyborn (approved):

ericneil-sanc (approved):

dannythedawger (approved):
lgtm

PR #738, opened 2026-09-18
feat(universal-router-sdk): update Universal Router v2.2.0 addresses

dianakocsis:
Updates `V2_2_0` to the routers deployed from universal-router tag `2.2.0` (`64027f3`), which carries the v4-periphery bump to `a7af5b34`.

| Chain | Address | Creation block |
| --- | --- | --- |
| Ethereum | `0xab863E752Bf67D8DCDD929EaAe9Be9dc83Fb3BbB` | 26006366 |
| Sepolia | `0x5093f1CDED83d99FfEd6602dA6260672ae16787c` | 11732372 |
| Ink | `0x8BbCAEb326AD590521a921C9e457Ee09F336B48c` | 56261823 |

Each was deployed with CREATE2 through the deterministic deployer using a zero salt, carries a 23,233-byte runtime matching the snapshot in Uniswap/universal-router#514, and is verified (Etherscan for mainnet and sepolia, Sourcify for Ink). Deployment PR: Uniswap/contracts#165.

## Ink's V2_1_1 entry is removed

Ink has no v2.1.1 router. The entry was an alias to the v2.2.0 deployment, justified in a comment on the grounds that v2.2.0 was a strict superset of v2.1.1 with "no change to any existing command/struct/behavior", so it served v2.1.1 callers byte-for-byte.

That no longer holds. The current v2.2.0 pins v4-periphery past [#564](https://github.com/Uniswap/v4-periphery/pull/564), which makes `V4Router` revert on exact-output partial fills. v2.1.1 does not. Keeping the alias would silently hand v2.1.1 callers a router that behaves differently on exact-output swaps.

Resolving `V2_1_1` on Ink now throws, consistent with how MegaETH already behaves. Callers wanting the 2.1 line on Ink have a real `V2_1_2` deployment at `0x661E93cca42AfacB172121EF892830cA3b70F08d`.

This is a breaking change for anyone currently calling `UNIVERSAL_ROUTER_ADDRESS(V2_1_1, 57073)` or `UNIVERSAL_ROUTER_CREATION_BLOCK(V2_1_1, 57073)`.

## Tests

Three tests updated as a consequence: the two Ink alias assertions now expect a throw, and the V2_1_2 coverage test treats Ink alongside MegaETH as a chain with 2.1.2 but no 2.1.1.

`bun run test` passes: 654 hardhat, 80 forge. Lint and build clean.


dannythedawger (approved):
lgtm

